Issue #131, opened 2014-11-07
node global installation process failure


jihoonl:
follow up from #127 

```
jihoonl@whoola:~$ sudo npm install -g roslib
[sudo] password for jihoonl: 
-
> roslib@0.9.0 preinstall /usr/local/lib/node_modules/roslib
> sh misc/install.sh

[sudo] password for nobody: 
Sorry, try again.
[sudo] password for nobody: 
Sorry, try again.
[sudo] password for nobody: 
Sorry, try again.
sudo: 3 incorrect password attempts
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
```

npm version 0.9.0 Testing.

**local installation**

> npm install roslib

 works fine.

**global installation**

> sudo npm install -g roslib

It seems like preinstall process causing problems.

* asking sudo for nobody
* node-canvas installation using curl command.

T045T:
This should be fixed with #178, since the installation no longer requires ```sudo```

PR #133, opened 2014-11-10
[Do not merge. under discussion] node global installation fix

jihoonl:
Discussion is going in #131


jihoonl:
Addressed by  https://github.com/RobotWebTools/roslibjs/issues/177

PR #151, opened 2015-02-04
0.11.0 Release

rctoris:
@jihoonl can you release nodejs version as well?


jihoonl:
The current version shoud be the same as old version. I will re-release when 0.13.0 get released.

rctoris:
I only reverted the TF client changes, there are a few other changes from 0.10.0 to 0.12.0 that remained.

PR #158, opened 2015-02-14
Index components of roslib

### LINKED ISSUE #157: Separate URDF functionality
I would like to use the urdf js code for some work that is unrelated to ROS. Do you think it would be possible to separate the ROS specific parts from the URDF parts? I had a quick look, and it seems to be not too much work. 
The urdf stuff could go into a new repo, which would also include all the three.js adapters in the visualisation library.

megawac:
This allows users to do

```js
> var Ros = require("roslib/src/core")

{ Ros: [Function: Ros],
  Topic: [Function: Topic],
  Message: [Function: Message],
  Param: [Function: Param],
  Service: [Function: Service],
  ServiceRequest: [Function: ServiceRequest],
  ServiceResponse: [Function: ServiceResponse]
}
```

Fixes #157

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ before_install:
   - cd ../
   - rosdep install --from-paths src --ignore-src --rosdistro hydro -y
   - cd /tmp/ws/roslibjs
+  - npm install -g npm
   - npm install -g grunt-cli karma-cli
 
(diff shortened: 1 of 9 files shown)


rctoris:
build failing even with re-trigger

PR #161, opened 2015-02-21
Add Urdf features: multiple visuals and material linking

jakobs:
These commits will add missing Urdf features. Note that this will break api compatibility, since the UrdfLink.visual param is missing. I am adding a PR to ros3djs as well.

### DIFF
--- a/src/urdf/UrdfMaterial.js
+++ b/src/urdf/UrdfMaterial.js
@@ -34,4 +34,14 @@ function UrdfMaterial(options) {
   }
 }
 
+UrdfMaterial.prototype.isLink = function() {
+  return this.color === null && this.textureFilename === null;
+};
+
+UrdfMaterial.prototype.assign = function(obj) {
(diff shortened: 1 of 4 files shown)


src/urdf/UrdfMaterial.js
```diff
@@ -34,4 +34,14 @@ function UrdfMaterial(options) {
...
 }
 
-module.exports = UrdfMaterial;
\ No newline at end of file
+UrdfMaterial.prototype.isLink = function() {
+  return this.color === null && this.textureFilename === null;
+};
+
+UrdfMaterial.prototype.assign = function(obj) {
+  for( var key in obj ) {
```
megawac: This can just be

```js
var assign = require("object-assign");

UrdfMaterial.prototype.assign = function(obj) {
  return assign(this, obj);
}
```

jakobs:
Thanks. Updated it.

PR #167, opened 2015-03-06
Indigo devel

psoetens:
This merge request adds a Quaternion function and fixes a misconception in a previous patch of Topic.js

According to the roslib websocket protocol queue_size and queue_length are different things and should not come from the same (default) parameter.


rctoris:
Is `queue_size` now completely unused?

psoetens:
Thanks for merging ! With the merge, both queue_size and queue_length are used, but the former for re-publishing and queue_size for pushing subscriptions to the web interface. It's part of the ROSBridge protocol, not my own inventions.

PR #171, opened 2015-03-18
Restore tf2_web_republisher service functionality

### LINKED ISSUE #152: New TF implementation not working
@T045T I released the new version of roslibjs with the new TF implementation and TFs are no longer working correctly. Looks like no TFs are handled correctly (see screenshot). 

untitled

Since this is a critical bug I'm going to revert and re-release!

T045T:
closes #152 - turns out ```trimLeft```was the culprit


src/tf/TFClient.js
```diff
@@ -45,15 +64,16 @@ function TFClient(options) {
...
  */
-TFClient.prototype.processFeedback = function(tf) {
+TFClient.prototype.processTFArray = function(tf) {
   var that = this;
   tf.transforms.forEach(function(transform) {
     var frameID = transform.child_frame_id;
-    if (frameID[0] === '/') {
+    if (frameID[0] === '/')
+    {
       frameID = frameID.substring(1);
```
megawac: this can't just be written as `frameID.trimLeft('/')` you're saying?

T045T: Yes. At least in the current version of Chrome, if you do
```
var trimmed = '/base_link'.trimLeft('/');
console.log(trimmed);
```
yields "/base_link".
The documentation [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/TrimLeft) doesn't mention ```trimLeft()``` taking any parameters, and advises against using it anyway.

Oddly enough, trimming seemed to work in some places but not others - I thought it safest to just use substring everywhere.

megawac: Hmm, would it be worthwhile to just import a node module for it?

https://www.npmjs.com/package/underscore.string
https://www.npmjs.com/package/lodash.trimleft

T045T: Since it's only used in this file, and only three times, I don't think it's necessary to import a module just for trimLeft. Also, I believe using built-in functions keeps the code a tiny bit more accessible.

What other string functions would be useful for us? If there's other areas where we can simplify code significantly, of course we should consider it.

PR #176, opened 2015-04-02
remove the (unused) DOMParser shim

Rayman:
The current DOMParser shim is not used, you can verify this by fixing the `package.json`:
```
"../util/DOMParser": "./src/util/shim/DOMParser.js"
```

The `xmlshim` package provides its own shim (which is already used in the roslibjs build). This pull request makes it more clear which shim is used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,8 +39,7 @@
     "aliases": {
       "canvas": "./src/util/shim/canvas.js",
       "eventemitter2": "./src/util/shim/EventEmitter2.js",
-      "ws": "./src/util/shim/WebSocket.js",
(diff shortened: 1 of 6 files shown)


megawac:
This will include the xmlshim in the browser no?

Rayman:
xmlshim provides the `browserify` field in the package.json. This will result in this file being used as shim: https://github.com/znerol/node-xmlshim/blob/master/browser.js

The roslibjs build will be almost exactly be the same because this file was already used. The only difference is that it is now more clear.

megawac:
Cool looks good in that case

Issue #177, opened 2015-04-07
Remove sudo from npm install


rctoris:
Currently `npm install` will run `misc/install.sh` which runs some sudo commands. This is unexpected behavior to many and breaks many automated scripts. I propose removing this from the npm install script and placing the instruction to run the script in the README instead.

megawac:
Related #133. Ideally we could detect if `libcario` is installed. If it isn't

a) install it (requiring sudo)
b) Error the build and give users instruction on how to install it

jihoonl:
I would suggest b) to resolve the issue.  This is issue of `node-canvas` not internal problem of `roslibjs`.

rctoris:
+1 to `b`

rctoris:
@jihoonl can you run a npm release to propagate this please?

jihoonl:
Done.

https://www.npmjs.com/package/roslib

PR #178, opened 2015-04-09
removed sudo from npm install and provides a check instead

rctoris:
addresses https://github.com/RobotWebTools/roslibjs/issues/177


misc/check.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# check for libcairo2
+if [ -f /usr/include/cairo/cairo.h ]; then
```
jihoonl: Would this approach works for other platforms e.g) redhat?

rctoris: My thought was `/usr/include` is pretty standard for installing headers. I can add an or check for `/usr/local/include/cairo/cairo.h` as well that might be more robust. We can add locations as they come up if needed. If there are not other objections I think we should merge and do an initial release of this patch since I know of several cases blocking on the the `sudo` problem.

jihoonl:
:+1:

PR #185, opened 2015-07-10
BSON Decoding of Binary Objects

DLu:
roslibjs component of https://github.com/RobotWebTools/rosbridge_suite/issues/169

Pairs with https://github.com/RobotWebTools/rosbridge_suite/pull/190

Current problem: BSON is a huge dependency. 

#### Original file sizes
```
-rw-rw-r-- 1 dlu dlu  65K Jul 10 17:08 roslib.js
-rw-rw-r-- 1 dlu dlu  27K Jul 10 17:09 roslib.min.js
```

#### New File Sizes with this PR
```
-rw-rw-r-- 1 dlu dlu 275K Jul 10 17:09 roslib.js
-rw-rw-r-- 1 dlu dlu 104K Jul 10 17:09 roslib.min.js
```

Is that a problem? Is there a better way?

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -2,7 +2,8 @@
   "globals": {
     "global": true,
     "Blob": true,
-    "FileReader": true
+    "FileReader": true,
+    "bson": true
   },
   "curly": true,
   "eqeqeq": true,
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,6 +1,8 @@
 {
   "globals": {
-    "global": true
+    "global": true,
+    "Blob": true,
+    "FileReader": true
   },
   "curly": true,
   "eqeqeq": true,
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
   },
   "dependencies": {
     "aliasify-patch": "^1.4.1",
+    "bson": "^0.4.7",
     "canvas": "1.1.3",
     "eventemitter2": "~0.4.13",
     "object-assign": "^2.0.0",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,6 @@
   },
   "dependencies": {
     "aliasify-patch": "^1.4.1",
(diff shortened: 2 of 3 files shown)


DLu:
Also, do you need to check in the updated built files? I didn't and I'm trying to tell if that's why the tests are failing.

rctoris:
You don't need to rebuild in on a PR. The tests seem to be failing because it can't load `Blob`. You should do a check to see if `Blob` is actually available https://github.com/RobotWebTools/roslibjs/pull/185/files#diff-49dcdae7a46fbde852bf41066508c9f7R111

DLu:
That was easier than I originally thought. I was trying to find the correct dependency to ensure Blob was defined. I think I did it correctly this time. Javascript is a weird language. 

The original question about the file size remains.

rctoris:
Is this pulling in all of bson into the build file? Can we make it a shim like `EventEmitter2` and such? I'm less familiar with the idea but I think this is the way to go.

https://github.com/RobotWebTools/roslibjs/tree/develop/src/util/shim
https://github.com/RobotWebTools/roslibjs/blob/develop/package.json#L38

DLu:
Ok, I've gotten it to the point where it works and builds. It will work correctly if you include the [https://raw.githubusercontent.com/mongodb/js-bson/master/browser_build/bson.js](bson header) and print an error message if you try to decode a bson message without the header. 

However, I don't think this is the right way to do it. I tried to make a shim, but had problems getting it to build with grunt AND work in the browser.

rctoris:
Since it is an optional feature we can make it optional like `ColladaLoader` is in `ros3djs`

megawac:
Are we not going to move this to an extension adapter, so the user could include if desired, to avoid the file size increase

```
require('roslib/adapters/bson')
```

Issue #160, opened 2015-02-21
node package fails to "build" with node v0.12.0


chrisl8:
The install is failing in node v0.12.0. It works with the older 0.10.35.
The failure appears to be related to the dependency on canvas 1.1.3

```
5520 verbose about to build /home/chrisl8/node_modules/roslib/node_modules/xmlshim
5521 verbose unlock done using /home/chrisl8/.npm/_locks/xmlshim-c8d70b8da5989174.lock for /home/chrisl8/node_modules/roslib/node_modules/xmlshim
5522 verbose about to build /home/chrisl8/node_modules/roslib
5523 verbose unlock done using /home/chrisl8/.npm/_locks/roslib-8c9bde6a3240ef46.lock for /home/chrisl8/node_modules/roslib
5524 verbose stack Error: canvas@1.1.3 install: `node-gyp rebuild`
5524 verbose stack Exit status 1
5524 verbose stack     at EventEmitter.<anonymous> (/usr/local/lib/node_modules/npm/lib/utils/lifecycle.js:213:16)
5524 verbose stack     at EventEmitter.emit (events.js:110:17)
5524 verbose stack     at ChildProcess.<anonymous> (/usr/local/lib/node_modules/npm/lib/utils/spawn.js:14:12)
5524 verbose stack     at ChildProcess.emit (events.js:110:17)
5524 verbose stack     at maybeClose (child_process.js:1008:16)
5524 verbose stack     at Process.ChildProcess._handle.onexit (child_process.js:1080:5)
5525 verbose pkgid canvas@1.1.3
5526 verbose cwd /home/chrisl8
5527 error Linux 3.13.0-45-generic
5528 error argv "/usr/local/bin/node" "/usr/local/bin/npm" "install" "roslib"
5529 error node v0.12.0
5530 error npm  v2.5.1
5531 error code ELIFECYCLE
5532 error canvas@1.1.3 install: `node-gyp rebuild`
5532 error Exit status 1
5533 error Failed at the canvas@1.1.3 install script 'node-gyp rebuild'.
5533 error This is most likely a problem with the canvas package,
5533 error not with npm itself.
5533 error Tell the author that this fails on your system:
5533 error     node-gyp rebuild
5533 error You can get their info via:
5533 error     npm owner ls canvas
5533 error There is likely additional logging output above.
5534 verbose exit [ 1, true ]
5535 verbose unbuild node_modules/roslib/node_modules/canvas
[...]

PR #189, opened 2015-08-20
Change dependencies to allow building with node 0.12.x

chrisl8:
This is to solve Issue #160 

I freely admit this may not be the best way to deal with the issue, but it works for me and has been working since July 1st, although there may be parts of the code that I don't use.

Feel free to reject this and implement a better solution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,11 +24,11 @@
   },
   "dependencies": {
     "aliasify-patch": "^1.4.1",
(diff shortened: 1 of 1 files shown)


jihoonl:
Addressed by #191

rctoris:
as well as https://github.com/RobotWebTools/roslibjs/pull/190

PR #190, opened 2015-08-28
replace xmlshim with xmldom

Rayman:
xmldom is a full JavaScript implementation of DOMParser with the same functionality. It is a lot more used and so I think it is more reliable (217.411 downloads in the last week vs 88).

As a result this project will no longer depend on libxmljs which makes it easier to install on platforms such as windows. All tests on both node and browser still run.

As a side effect I had to reintroduce `shim/xmldom.js` to keep the build size small. I also replaced the xpath `evaluate` call with `documentElement` because we no longer have that. Could that become a problem?

### DIFF
--- a/test/urdf.test.js
+++ b/test/urdf.test.js
@@ -1,7 +1,7 @@
 var expect = require('chai').expect;
 var ROSLIB = require('..');
 
-var DOMParser = require('xmlshim').DOMParser;
+var DOMParser = require('xmldom').DOMParser;
 // See https://developer.mozilla.org/docs/XPathResult#Constants
 var XPATH_FIRST_ORDERED_NODE_TYPE = 9;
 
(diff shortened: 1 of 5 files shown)


test/urdf.test.js
```diff
@@ -114,7 +114,7 @@ describe('URDF', function() {
     it('is ignorant to the xml node', function(){
       var parser = new DOMParser();
       var xml = parser.parseFromString(sample_urdf(), 'text/xml');
-      var robotXml = xml.evaluate('//robot', xml, null, XPATH_FIRST_ORDERED_NODE_TYPE, null).singleNodeValue;
+      var robotXml = xml.documentElement;
```
megawac: Nice!!! Great stuff @Rayman

PR #194, opened 2015-09-01
Remove canvas dependency

Rayman:
The canvas dependency makes this library difficult to install on for example travis and windows. This PR removes that dependency such that no native Node.js extensions are needed (no more libcairo2-dev libjpeg8-dev libpango1.0-dev libgif-dev build-essential g++, just JavaScript code).

I modified the logic a bit but I think I tested all the code paths. Maybe we need CI tests to check the library with BSON and/or png compression.

- in a browser a canvas element will (still) be used for decoding pngs (no
  external dependencies needed)
- in node it will use the pngparse module which will use the native zlib
  module to inflate the data


package.json
```diff
@@ -40,7 +41,8 @@
       "canvas": "./src/util/shim/canvas.js",
       "eventemitter2": "./src/util/shim/EventEmitter2.js",
       "ws": "./src/util/shim/WebSocket.js",
-      "xmldom": "./src/util/shim/xmldom.js"
+      "xmldom": "./src/util/shim/xmldom.js",
+      "../util/decompressPng": "./src/util/shim/decompressPng"
```
megawac: should it be ./src/util...

Rayman: That's what I also thought but this is the way aliasify works apparently :disappointed: We could rewrite it using the [browser field of the package.json](https://github.com/substack/browserify-handbook#browser-field) and not use aliasify.

Rayman:
Fixed the merge conflict and removed aliasify so that it is more clear how the shims work.

Issue #114, opened 2014-10-02
Mysterious Test Failures


T045T:
There seems to be something strange going on with the automated tests. Maybe a race condition?

[This line](test/examples/fibonacci.example.js#40) could cause one, but why would it, with the fibonacci server inserting pauses in between feedback updates?

rctoris:
Are these still failing for people?

rctoris:
Closing for inactivity / better stability noticed in travis builds

PR #196, opened 2015-09-10
Test more node versions in travis

Rayman:
I added some more node versions to test for and cleaned the travis script. The current version of node is v4.0.0 and roslib doesn't work on that due to an issue in Automattic/node-canvas#622. We could fix this with #194.

Version 1 and 2 are versions of io.js but apparently roslib works fine on there. Version 3 was the 'alpha' of node v4.


rctoris:
Looks like the builds are failing

Rayman:
I rebased and it looks like travis still fails but I'm unable to reproduce that issue on my machine. During an other build it looked like issue #114 occured again... https://travis-ci.org/Rayman/roslibjs/builds/80210067. I'll continue looking for a solution.

Rayman:
I added the dependencies for chai and mocha and now everything works again.

Rayman:
Rebase to fix a merge conflic in package.json

PR #202, opened 2015-10-14
Add options to omit the feedback, status or result channel of ActionClient.

commaster90:
Omitting certain channels of the action client will make the action protocol incomplete. However, this can reduce the network load of the websocket and the action protocol will still work in certain use cases.


rctoris:
Nice idea, I like it.

commaster90:
Thanks.

PR #208, opened 2016-02-04
Update roslibjs to advertise Services

chris-smith:
Updates roslib so that it can advertise services in addition to calling them.

Example Usage
```
let addTwoInts = new roslib.Service({
    ros: rosConn,
    name: '/add_two_ints',
    serviceType: 'beginner_tutorials/AddTwoInts'
});

addTwoInts.advertise((req, resp) => {
    resp.sum = req.a + req.b
    return true;
});
```

### DIFF
--- a/src/core/Service.js
+++ b/src/core/Service.js
@@ -4,6 +4,8 @@
  */
 
 var ServiceResponse = require('./ServiceResponse');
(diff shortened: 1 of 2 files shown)


rctoris:
:+1:

reinzor:
:+1:

PR #220, opened 2016-03-28
An action listener for all actions not just for goals created by client

FuzzAU:
I had the need to listen for arbitrary goal related messages, not just for goals created in my JS client. I wish to monitor goals and display feedback and status for goals created by another client.

As such I built this action listener (based on the existing action client) that let me subscribe to the topics directly rather than filtering based on the goals that I'd created.

I wasn't sure if there is apetite for this feature, so thought I'd submit the pull request to seek interest.


rctoris:
Cool idea!

PR #224, opened 2016-04-15
add RTCPeerConnection as Ros transport library

### LINKED ISSUE #223: allow Ros.connect to use an RTCPeerConnection/RTCDataChannel

aisch:
closes #223

not sure if this is best way to add this support to roslibjs but with this patch i was able to connect a [RTCDataChannel](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel) to a `rosbridge` like e.g.

```js
ros.transportLibrary = peer_connection;
ros.transportOptions = {
  reliable: false,
  ordered: false,
};
ros.connect('rosbridge');
```


aisch:
@rctoris thanks!

Interpause:
<https://github.com/RobotWebTools/rosbridge_suite> does not seem to have any ICE server implementation for WebRTC, how is this feature intended to be used? I am planning to use WebRTC's DataChannel as well, and would not mind attempting to follow the rosbridge protocol, but I am unclear what the current implementation status is.

PR #231, opened 2016-07-18
Updated all dependency libraries to the latest

dantwinkler:
The major bug that this fixes a nasty ws bug.  ws version 1.1.1 fixes this issue along with other more serious security issues.  All unit tests successfully pass with the updated packages.

https://github.com/websockets/ws/issues/778


jihoonl:
:+1:

PR #221, opened 2016-04-07
Upgraded EventEmitter2 and removed shim

eadlam:
I upgraded the EventEmitter2 version in package.json and removed the shim. This allows roslibjs to be used in webpack projects.

PR #240, opened 2016-11-17
Upgrade to EventEmitter 2.2.0

jbcpollak:
This is an update of https://github.com/RobotWebTools/roslibjs/pull/221

- It builds on @eadlam's changes and upgrades EventEmitter2 to `~2.2.0`.
- In addition, I replaced the deprecated function `utils.debug` with `console.error`.

I did the following:

* updated README.md (EventEmitter isn't the only dependency, so I make the comments generic)
* Ran tests (they pass)

_Note_:

* There are failing `test-examples`, but the same tests fail on the main `develop` branch.
* I cannot update the cdn with the latest version of EventEmitter, I don't know how that works.

I should add this has been experimentally tested with our own code and works fine.


jihoonl:
Thanks. and sorry for huge delay.

PR #245, opened 2016-12-30
Fix getTopics bug

NickZ:
If a user did not pass in a failedCallback into getTopics(), it would only return the "topics" array from the result, rather than the full result that included both the "topics" and the "types" array.


jihoonl:
:+1:

PR #250, opened 2017-02-25
Support status level messages

sevenbitbyte:
This patch adds the ability to change the global and operation status reporting level as well as receiving status messages from the server.

https://github.com/RobotWebTools/rosbridge_suite/blob/develop/ROSBRIDGE_PROTOCOL.md#32-status-messages


jihoonl:
:+1:

PR #265, opened 2017-10-16
Add Release Todo document

jihoonl:
Hello @RobotWebTools/maintainers,

@all, I roughly wrote down todos to prepare a new version release with some discussion points. Please take a look and leave a comment. Correcting my english is also very welcome. And does anyone experience with bowers?

@nuclearsandwich I think some steps are automatable. e.g) (update changelog, version bump and tagging), (Releasing in npm and bowers).

@sevenbitbyte having a scoped package approach seems a good idea to manage multiple modules under the same namespace. But, I would like to continue to keep roslibjs in global scope because it has  been served in global scope for ages already. Do you know any good way of managing a module for both global and org scoped?

@jihoonl will fill out CDN part after releasing npm module.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,23 +4,23 @@ Robot Web Tools Module Release TODOs
 This document describes TODOs and checklists in order to release
 Robot Web Tool javascript modules([roslibjs](https://github.com/RobotWebTools/roslibjs), [ros2djs](https://github.com/RobotWebTools/ros2djs), [ros3djs](https://github.com/RobotWebTools/ros3djs)).
 
-#### Make sure that the releasing module is compatible with other RWT modules.
+### Make sure that the releasing module is compatible with other RWT modules.
 
 use travis to check the compatibility?
 
-#### Generate CHANGELOG using [github-changes](https://github.com/lalitkapoor/github-changes). 
+### Generate CHANGELOG using [github-changes](https://github.com/lalitkapoor/github-changes). 
 
 ```
 github-changes -o RobotWebTools -r roslibjs --only-pulls --use-commit-body -a -b develop
 ```
 
-#### Bump a new version
+### Bump a new version
 
 * Remove a SNAPSHOT postfix from package.json and bowers.json
 * Mark *upcoming* in CHAGELOG.md as the new release version
 * Tag the version
 
-#### Release
+### Release
 
 **NPM**
 
@@ -34,17 +34,17 @@ Can it be automated too?
 
 TODO(@jihoonl) Fill it after release
 
-#### Update jsdocs in Robot Web Tools website
+### Update jsdocs in Robot Web Tools website
 
 [Here](https://github.com/RobotWebTools/robotwebtools.github.io/tree/master/jsdoc).
 Check if docs can be hosted directly from its own repository.
 
-#### Sync `develop` branch with `master`
+### Sync `develop` branch with `master`
 
 `Master` branch should represent the latest release. 
 * Create a PR against `master` from `develop`
 * Do *Rebase and merge* to have the same history as `develop` branch
 
-#### Prepare `develop` branch for a new version
+### Prepare `develop` branch for a new version
 
 Bump a version in package.json and bowers.json and add `-SNAPSHOT` postfix.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,23 +4,23 @@ Robot Web Tools Module Release TODOs
 This document describes TODOs and checklists in order to release
 Robot Web Tool javascript modules([roslibjs](https://github.com/RobotWebTools/roslibjs), [ros2djs](https://github.com/RobotWebTools/ros2djs), [ros3djs](https://github.com/RobotWebTools/ros3djs)).
 
-### Make sure that the releasing module is compatible with other RWT modules.
+### 0. Make sure that the releasing module is compatible with other RWT modules.
 
 use travis to check the compatibility?
 
-### Generate CHANGELOG using [github-changes](https://github.com/lalitkapoor/github-changes). 
+### 1. Generate CHANGELOG using [github-changes](https://github.com/lalitkapoor/github-changes). 
 
 ```
 github-changes -o RobotWebTools -r roslibjs --only-pulls --use-commit-body -a -b develop
 ```
 
-### Bump a new version
+### 2. Bump a new version
 
 * Remove a SNAPSHOT postfix from package.json and bowers.json
 * Mark *upcoming* in CHAGELOG.md as the new release version
 * Tag the version
 
-### Release
+### 3. Release modules
 
 **NPM**
 
@@ -34,17 +34,17 @@ Can it be automated too?
 
 TODO(@jihoonl) Fill it after release
 
-### Update jsdocs in Robot Web Tools website
+### 4. Update jsdocs in Robot Web Tools website
 
 [Here](https://github.com/RobotWebTools/robotwebtools.github.io/tree/master/jsdoc).
 Check if docs can be hosted directly from its own repository.
 
-### Sync `develop` branch with `master`
+### 5. Sync `develop` branch with `master`
 
 `Master` branch should represent the latest release. 
 * Create a PR against `master` from `develop`
 * Do *Rebase and merge* to have the same history as `develop` branch
 
-### Prepare `develop` branch for a new version
+### 6. Prepare `develop` branch for a new version
 
 Bump a version in package.json and bowers.json and add `-SNAPSHOT` postfix.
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,50 @@
...
+* Mark *upcoming* in CHAGELOG.md as the new release version
+* Tag the version
+
+### 3. Release modules
+
+**NPM**
+
+Keep all modules under @robot-web-tools scope? or in global scope? What are pros and cons?
+
+**Bowers**
```
nuclearsandwich: As far as I know bower is a thin wrapper on Git itself. Once a package is [registered](https://bower.io/docs/creating-packages/#register) and has a valid bower.json all you need to do to update is create and push a new git tag.

nuclearsandwich:
I noticed that git-tagging the the release isn't included in the checklist. It could be added to the bower release instructions but it's also worth doing separately so may be its own checklist item.

sevenbitbyte:
@jihoonl There does appear to be an approach for tagging multiple maintainers using [`npm owner add`](https://docs.npmjs.com/cli/owner). It also does not appear to require a new release to be updated either so maybe thats a good approach for roslib.

RELEASE.md
sevenbitbyte: Bower is kinda dying right now as the alternatives have won out over the years (sources [official1](https://bower.io/blog/2017/how-to-migrate-away-from-bower/), [official 2](https://github.com/bower/bower/blob/master/README.md#bower---a-package-manager-for-the-web), [reddit 3](https://www.reddit.com/r/javascript/comments/5ls30m/is_bower_dead_should_we_start_using_other/)). The current approaches people have been using is to just use NPM and not do much else special for frontend vs backend nodejs package management. I have very successfully used npm+[webpack](https://webpack.github.io/) or npm+[browserify](http://browserify.org/) with all of the RWT packages. [Yarn](https://yarnpkg.com/en/) is also a common alternative.

sevenbitbyte:
## Scoping

On the topic of global vs org scoped packages, I've just noticed that we can have an org without requiring the name to be scoped. That is we can do either while using an org so both global (`roslib`, `ros3djs`, `ros2djs`) or scoped (`@robot-web-tools/{package}`) is allowed under an org. The benefit of the org being that it allows us to manage maintainers easily through the web interface with granular permissions on a per team/package basis. The org also giving our users a very clear line of communication via the org page itself. In mature projects like [Google's Firebase team for instance](https://www.npmjs.com/org/firebase) it gives consuming developers a single place to look for all of the packages. So it seems for legacy reasons we should simply add the npm org as a co-owner using the `npm owner add` command and continue using global names so long as all of our package names are available.

## Bower

Bower is dying. We should do one final release and cut bower support.

nuclearsandwich:
> Bower is dying. We should do one final release and cut bower support.

While Bower does recommend folks use yarn or webpack, it is still maintained and given the relatively low effort (keeping the bower.json present and up-to-date, plus tagging which we do anyway) is there any reason not to continue supporting it as long as bower itself is maintained and it remains low effort to do so?

sevenbitbyte:
@nuclearsandwich You're right, its too soon to remove altogether. My biggest concern is upstream libraries getting stagnate(just checked they all look good for now) and not wanting to lead new comers down an unsupported path accidentally. So we should just be aware

jihoonl:
**Bower**
Thanks for the inputs. It would be ok to keep the Bower release since we tag versions anyway in Git. I was inferring git-tagging from _Tag the version in 2. Version Bump_.

**Scoped packages in npm**
@sevenbitbyte thanks for the detailed explanation. It would help us to manage multiple modules as a team. I propose to release all three modules under org scope, then release in global scope after testing. We may use org scoped version as unstable release and push frequently. How do you think?

**Yarn**
I quickly read some articles regarding _yarn_. It seems like yarn is introduced to fix headaches come from dependency management. Looks interesting but we can take a look later independently from this release process review.

jihoonl:
@nuclearsandwich Could you help me to automate chagelogging, git-tagging, and releasing in npm? What's your npmjs id? I will add you in organization.

@sevenbitbyte I will replace the global version of roslibjs when org-scoped version of roslibjs is ready.

@adolfo-rt , @viktorku How do you manage the versioning and releasing of your modules? Could you comment on the current draft?

nuclearsandwich:
> My biggest concern is upstream libraries getting stagnate

This is a valid concern @sevenbitbyte and one I hadn't considered! What procedure did you use to check the upstream dependency availability on bower? If possible we should script it and add it to the bower release checklist. If one of our dependencies isn't available or is out of date on bower we can ask them to release or just drop bower support ourselves.

Should we add some kind of deprecation notice to the description in the bower.json? Or do we think that all of bower is essentially "deprecated" and that end users likely know this?

jihoonl:
> This is a valid concern @sevenbitbyte and one I hadn't considered! What procedure did you use to check the upstream dependency availability on bower? If possible we should script it and add it to the bower release checklist. If one of our dependencies isn't available or is out of date on bower we can ask them to release or just drop bower support ourselves.
> Should we add some kind of deprecation notice to the description in the bower.json? Or do we think that all of bower is essentially "deprecated" and that end users likely know this?

[Bower officially recommends moving away from itself](https://bower.io/blog/2017/how-to-migrate-away-from-bower/). So, I don't see any big reason to put extra effort to maintain it further. I will add a deprecation warning in bower.json descrption field. Then, we can completely drop the bower support maybe next year?  If somebody complains breakage in bower via github issue, we can guide them to use Webpack or browserify.

nuclearsandwich:
> @nuclearsandwich Could you help me to automate chagelogging, git-tagging, and releasing in npm? What's your npmjs id? I will add you in organization.

I'm [nuclearsandwich](http://npmjs.org/~nuclearsandwich) on npm as well. I'll take a first pass at automating release this week or weekend.

RELEASE.md
jihoonl: As we are deprecating bower support, I will remove this section. Yarn can be added later.

jihoonl:
@sevenbitbyte I just tried to release the version 0.19.0 under org scope. But, it got into global namespace again. Can you guide me how to release module under org?

```
# Tag 0.19.0 and push into github
> npm config set scope robot-web-tools
> npm adduser # add myself
> npm publish
```
This is what I did. What step am I missing?

nuclearsandwich:
Thanks to @jihoonl we've marked bower support as "deprecated" in the bower.json description. https://github.com/RobotWebTools/roslibjs/pull/271

nuclearsandwich:
> @sevenbitbyte I just tried to release the version 0.19.0 under org scope. But, it got into global namespace again. Can you guide me how to release module under org?

My impression from @sevenbitbyte's comment was that the package would be published both globally and into the `@robot-web-tools` org scope once the organization has been added as an owner on the roslib npm package.

Looking at [roslib on npm](https://www.npmjs.com/package/roslib) it appears that you are the only current owner.

So running

```
npm owner add robot-web-tools roslib
```

should add it to the organization. I'm still reading the documentation though so that may not be the full story.

jihoonl:
For a note, this command helped to add robot-web-tools:devleopers to have co-ownerwhip of `roslib` 

```
> npm access grant read-write robot-web-tools:developers roslib
```

I am stilll failing to publish `roslib` in `robot-web-tools` scope though.

nuclearsandwich:
> I am stilll failing to publish roslib in robot-web-tools scope though.

Does this step require the ownership that @sevenbitbyte suggested? I don't know for certain. I have never been part of an organization on npm.

sevenbitbyte:
Hi @jihoonl and sorry about the slow update.

It looks like `roslib` is now successfully listed on the organization page here: https://www.npmjs.com/org/robot-web-tools

After reviewing a bit I think to scope the package you simply need to publish after changing the package name to `@robot-web-tools/roslib` rather than the current `roslib`. It looks like I did this slightly incorrectly over in ros3djs so I just posted a patch for fix that: https://github.com/RobotWebTools/ros3djs/pull/198

jihoonl:
I think we need to move carefully here because changing package name is something I would like to avoid unless there is a critical reason. 

I reread all linked documents about scoped packages and [firebase](https://www.npmjs.com/org/firebase) example. It looks like I misunderstood org-scoped package and mixed two discussions here. 1) use npm org to manage RWT modules. 2) release RWT as scoped packages(rename modules as `@robot-web-tools/<XXX>` and release into npm).

I agree with 1) managing modules using npm org. Like @sevenbitbyte mentioned, it will help us to manage multiple packages as a team. But, I need to disagree with 2) releasing RWT as scoped packages. `roslib` has served as a global scoped package for many years already and it looks like adding scoped package release can be an extra burden to me.

To conclude, I would like to manage RWT modules under [robot-web-tools org](https://www.npmjs.com/org/robot-web-tools), but release them as global scoped packages(`roslib`, `ros2d`, and `ros3d`). And we can create something like `robot-web-tools` package to install all packages in org. How about this?

jihoonl:
ping. Any more comments? @sevenbitbyte

PR #271, opened 2017-10-30
Add deprecation warning in bower.json description

jihoonl:
As we discussed in #265, deprecation warning is added in bower module descrption.
@nuclearsandwich review and merge please.


nuclearsandwich:
I removed a word from the start of the description sentence that didn't fit with the rest of the sentence. With that change this looks good to me.

Good idea to remove the suggestion of using bower in the README as well 👍

nuclearsandwich:
@jihoonl before I do something incorrect. You would like me to handle this merge correct?

jihoonl:
Yes correct go ahead

nuclearsandwich:
🎆

PR #277, opened 2017-12-19
Use https instead of http in examples and links

jihoonl:


jihoonl:
@T045T please use "squash and merge" when you merge. thanks.

T045T:
Looks good, but just to be 100% sure, I restarted the failed Travis job (looks like that failed because of a network problem).

PR #280, opened 2018-01-08
Add post processing logic to update built module by travis

jihoonl:
* and updates travis config to use trusty


jihoonl:
Well..not working.

PR #282, opened 2018-01-22
Add the type of service when executing 'call_service'

minggangw:
We want to reuse the roslibjs as the front-end JavaScript library by
leveraging the ros2-web-bridge in ROS2. But the current roslibjs doesn't
transmit the service type when calling a service, which is a necessary
element when sending the request in ROS2.

This patch adds this information to make it compatible with ROS2.

### DIFF
--- a/src/core/Service.js
+++ b/src/core/Service.js
@@ -59,6 +59,7 @@ Service.prototype.callService = function(request, callback, failedCallback) {
     op : 'call_service',
     id : serviceCallId,
     service : this.name,
+    type: this.serviceType,
     args : request
   };
   this.ros.callOnConnection(call);


minggangw:
Hi @jihoonl,

When I was developing the [ros2-web-bridge](https://github.com/RobotWebTools/ros2-web-bridge), I found this [problem](https://github.com/RobotWebTools/ros2-web-bridge#known-issues). Although I had a very tricky workaround, which lets the code odd, I think it would better change the library.

I am not sure whether it's proper to do it in order to make roslibjs workable in ROS2, so please help me review this patch. I hope the roslibjs can be reused in ROS2 by adopting some minor changes. Thanks!

PR #286, opened 2018-04-13
update shim files to use window instead of global

jorgenfb:
The files in /shim is only used in browsers (remapped in the browser
field of package.json) when using build tools. Browsers do not have the
global propery which is a node thing. Browsers instead have the global
namespace window.

This commit solve my problem of importing roslib in my application and
building with webpack 4 without having webpack shim global = window.

Note: updated Gruntfile.js to have jshint read the config file from the
closes directory/parent directory and added a new .jshintrc to /shim to
lint the files there with a browser environment in mind.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -18,7 +18,7 @@ module.exports = function(grunt) {
     },
     jshint: {
       options: {
-        jshintrc: '.jshintrc'
+        jshintrc: true
       },
       files: [
         './Gruntfile.js',
--- a/src/util/shim/.jshintrc
+++ b/src/util/shim/.jshintrc
@@ -1,0 +1,24 @@
+{
+  "globals": {
+    "Blob": true,
+    "FileReader": true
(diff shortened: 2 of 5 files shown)


jorgenfb:
Sure! Build artifacts removed.

yokyj:
hi guys
to improve performance,I want roslib to suscribe topics in web worker
but there is not objetc window in worker field so it crashed

jorgenfb:
Hi! I do not think this library can be used in a worker out of the box. Mainly because it references several global objects not available in a worker. For instance DOMParser, DOMImplementation, XMLParser, Canvas and Image. I suggest that you open a new issue regarding support for usage in web workers.

mvollrath:
@yokyj there is now a feature to run the WebSocket in a Worker, though decoding still happens in the main thread.  See https://github.com/RobotWebTools/roslibjs/pull/317

PR #303, opened 2018-11-05
Add support for CBOR encoding

mvollrath:
Goes with https://github.com/RobotWebTools/rosbridge_suite/pull/364

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "pngparse": "^2.0.1",
     "ws": "^1.1.1",
     "xmldom": "^0.1.19",
+    "cbor-js": "^0.1.0",
     "socket.io": "1.4.8"
   },
   "directories": {
--- a/src/core/Ros.js
+++ b/src/core/Ros.js
@@ -70,7 +70,9 @@ Ros.prototype.connect = function(url) {
   } else if (this.transportLibrary.constructor.name === 'RTCPeerConnection') {
     this.socket = assign(this.transportLibrary.createDataChannel(url, this.transportOptions), socketAdapter(this));
   }else {
-    this.socket = assign(new WebSocket(url), socketAdapter(this));
+    var sock = new WebSocket(url);
+    sock.binaryType = 'arraybuffer';
+    this.socket = assign(sock, socketAdapter(this));
   }
 
 };
--- a/src/core/SocketAdapter.js
+++ b/src/core/SocketAdapter.js
@@ -9,6 +9,8 @@
 'use strict';
 
 var decompressPng = require('../util/decompressPng');
+var CBOR = require('cbor-js');
+var typedArrayTagger = require('../util/cborTypedArrayTags');
 var WebSocket = require('ws');
 var BSON = null;
 if(typeof bson !== 'undefined'){
@@ -106,6 +108,9 @@ function SocketAdapter(client) {
         decodeBSON(data.data, function (message) {
           handlePng(message, handleMessage);
         });
+      } else if (data.data instanceof ArrayBuffer) {
+        var decoded = CBOR.decode(data.data, typedArrayTagger);
+        handleMessage(decoded);
       } else {
         var message = JSON.parse(typeof data === 'string' ? data : data.data);
         handlePng(message, handleMessage);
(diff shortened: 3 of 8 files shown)


mvollrath:
I intend to add some tests this week.

build/roslib.js
```diff
@@ -1,4 +1,412 @@
...
+if (typeof define === "function" && define.amd)
+  define("cbor/cbor", obj);
+else if (typeof module !== 'undefined' && module.exports)
+  module.exports = obj;
+else if (!global.CBOR)
+  global.CBOR = obj;
+
+})(this);
+
+},{}],2:[function(require,module,exports){
```
viktorku: This chunk of code should not appear solely in this file, `roslib.js` is the result of all packed non-minified modules in the repo.

viktorku:
> Are you saying I should remove all changes to build/ from this PR?

Yes, the `build` directory is not supposed to have any manual changes - that's the result of packing all modules in `src`. You can freely add new modules where they seem appropriate, in `core` and/or `util` in your case mostly I think.

mvollrath:
Added my tests and removed the build cruft.

jihoonl:
We might want to remove build directory as we have seen this minified version drags some PRs down.

subvertallchris:
I know I'm coming in here way late but just felt obligated to say that I really don't love the idea of adding a dependency ([cbor-js](https://github.com/paroga/cbor-js)) that hasn't been updated or seen a closed issue since 2016 and has [benchmarks](https://github.com/paroga/cbor-js/issues/11) to indicate that it's significantly slower than JSON. 

I'm not suggesting that this be immediately removed if people are using it but I'd love to know what kind of adoption it really has. In the meantime, is it worth considering some kind of plugin system that would let someone opt in to this or other libraries like it?

mvollrath:
> I know I'm coming in here way late but just felt obligated to say that I really don't love the idea of adding a dependency ([cbor-js](https://github.com/paroga/cbor-js)) that hasn't been updated or seen a closed issue since 2016 and has [benchmarks](https://github.com/paroga/cbor-js/issues/11) to indicate that it's significantly slower than JSON.

If you have a favorite CBOR library, feel free to drop it in.  I tried all of them and decided to use this one.  CBOR is still a new protocol and I expect by the time there is a maintenance problem with this library a better one will be available.

The performance difference depends a lot on what kind of data is being deserialized, for all the types of data this patch was intended to optimize (large numeric arrays and untyped binary data e.g. PointCloud2 buffers) CBOR is significantly faster than JSON and its output (typed arrays) is more useful.  This difference is described in the rosbridge documentation.  This is why CBOR is not the default encoding for rosbridge.

subvertallchris:
My point is more that I don’t know why everyone who uses roslibjs or
rodbridge should be required to download a CBOR dependency at all. While I
appreciate your interest in wanting to use a new data format, I’m not sure
it’s mainstream or mature enough to justify its inclusion in the libraries.
A plug-in system that makes it easy to use this and things like it would
make it easier to try things like this.

mvollrath:
I was interested in making rosbridge fast enough so that people can send all types of data over it instead of having to develop special socket servers or republishers for things like point clouds.  The people using it seem to be happy with it.  I'd rather not make things more complicated for them unless there's a really good reason for it.  From where I'm sitting, the CBOR feature only makes up about 10% of the total LoC&C for this client project.  The runtime footprint is [one `if` condition per message](https://github.com/RobotWebTools/roslibjs/blob/969f568ceeb15600ebea1a5047bca5b34b7e58ad/src/core/SocketAdapter.js#L106-L116) if you never invoke it.  That doesn't seem like a significant burden for anybody, considering what it does for people who need it.

subvertallchris:
I hear what you're saying. I apologize if this comes off as rude or dismissive of your interest in pushing things forward because I totally respect the work that went into all of this. I just think that it's worth being sensitive about which dependencies make it into a project and why they're added, especially if new dependencies are on unmaintained libraries. I'm not suggesting we roll back the clock and pull this out -- the ship has sailed on that. I was only chiming in so this could be considered for future PRs.

PR #304, opened 2018-11-15
Prevent multiple web socket connections being opened.

achim-k:
When `connect(...)` was called on an already connected `Ros` instance, a new web socket connection was initiated and the previous one not closed. This patch makes sure that we only create a new connection if there is none yet or the current one has been closed.

Note that this is only done for `WebSocket` connection here. Should ideally be implemented also for the other transports.

### DIFF
--- a/src/core/Ros.js
+++ b/src/core/Ros.js
@@ -69,8 +69,10 @@ Ros.prototype.connect = function(url) {
     this.socket.on('error', this.socket.onerror);
   } else if (this.transportLibrary.constructor.name === 'RTCPeerConnection') {
     this.socket = assign(this.transportLibrary.createDataChannel(url, this.transportOptions), socketAdapter(this));
-  }else {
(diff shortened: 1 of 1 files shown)


src/core/Ros.js
```diff
@@ -69,8 +69,10 @@ Ros.prototype.connect = function(url) {
     this.socket.on('error', this.socket.onerror);
   } else if (this.transportLibrary.constructor.name === 'RTCPeerConnection') {
     this.socket = assign(this.transportLibrary.createDataChannel(url, this.transportOptions), socketAdapter(this));
-  }else {
-    this.socket = assign(new WebSocket(url), socketAdapter(this));
+  } else {
+    if (!this.socket || this.socket.readyState == WebSocket.CLOSED) {
```
viktorku: Can you just make the equality `===`?

PR #305, opened 2018-11-27
Dependency version updated to the latest for 'ws'

imutkarshpatil:
Existing version of '_ws_' is vulnerable for DoS attacks with **high severity**. Version number changed to the latest where these are patched.

As mentioned here:
> https://nodesecurity.io/advisories/550
> https://nodesecurity.io/advisories/120


viktorku:
This is the same as a commit in #307, so I would rather merge that one since it upgrades `socket.io` as well. I'll merge that one in as soon as Travis succeeds, closing this one.

PR #307, opened 2018-12-30
upgrade dependencies

DorianScholz:
to avoid security vulnerabilities


package.json
```diff
@@ -36,10 +36,10 @@
     "eventemitter2": "~2.2.0",
     "object-assign": "^4.0.1",
     "pngparse": "^2.0.1",
-    "ws": "^1.1.1",
+    "ws": "^3.3.1",
     "xmldom": "^0.1.19",
     "cbor-js": "^0.1.0",
-    "socket.io": "1.4.8"
+    "socket.io": "2.2.0"
```
viktorku: Travis fails due to old nodejs 4.2.6 on the build/test image that doesn't support ES6 syntax (which socket.io 2.2.0 apparently uses).

chrisl8:
We need this update. What can I/we do to help make this happen?
I apologize if this is rude, as I am not sure what the procedure is here.
Thank you!

TobiasMorell:
I'm currently building a website using roslibjs and [preact](https://preactjs.com/). When building for release mode, it fails because roslibjs uses an outdated version of socket.io.

For now I was able to solve the issue by simply deleting socket.io from roslib/node_modules and replacing it with a newer version, as so:
```
rm -r node_modules/roslib/node_modules/socket.io
cp node_modules/socket.io node_modules/roslib/node_modules/socket.io
```
I'm fully aware that it's a bad solution, but while we're waiting for a fix, it's atleast something.

Anyways, I suggest updating to a newer node version and using babel to transpile to old js code.

TobiasMorell:
In case anyone else is experiencing the same issues as me (i.e. preact), I have devised a somewhat better solution to the problem. Run the following bash command before building:
`sed -i '26i\if(typeof window !== "undefined")\' node_modules/roslib/node_modules/socket.io/lib/index.js`

viktorku:
Thanks!

PR #308, opened 2019-01-01
Use melodic-ros-core image for building

viktorku:


viktorku:
Initially I thought this would bring a newer version of Node, but since #309 has been merged, I'll close this.

PR #309, opened 2019-01-01
Use LTS node v10

viktorku:
Do we want to still keep using kinetic?


jihoonl:
The majority are still using Kinetic. So I would prefer to keep the support of kinetic version. However, I am not sure whether we should keep node version as `4.x.x`. @mvollrath  @viktorku. I think you guys are the most experienced users and maintainers here. Feel free to move on if you both agree.

PR #317, opened 2019-01-29
Add WebSocket in WebWorker transport

mvollrath:
Decouples the application from WebSocket backpressure.

A client-side solution for https://github.com/RobotWebTools/rosbridge_suite/issues/212

To summarize, when you're rendering graphics (or doing anything expensive) in an application using ROSLIB, if the rendering takes to much time it causes backpressure on the WebSocket which eventually causes OOM on the rosbridge server.  This transport decouples the WebSocket from the main thread, introducing marginal cost to copy from the Worker to the main thread.  It should crash the client before crashing the server.


dirk-thomas:
@jihoonl I would be great if you could review this when you have a change :bowing_man:

jihoonl:
@dirk-thomas  Can you confirm it works with your usecase? Then, we can just proceed to merge.

jihoonl:
@viktorku Would it break your case?

viktorku:
It seems to be backwards compatible (doesn't remove the non-worker websocket comms) so I'm pretty sure this can go in.

Just one question for my understanding: you mention that it should crash the client before crashing the server. How does decoupling the socket data transmission to a separate thread/worker on the client achieves this?

mvollrath:
Regarding compatibility, `workersocket` is not the default `transportLibrary`, you have to request it when you instantiate `ROSLIB.Ros`.  This was done because `workerSocket` does not work in Node, only in a browser, making it the default would break the tests and probably some applications.  I made an additional browser test just for `workerSocket`.

About how this protects the server, it prevents slow applications using ROSLIB from being slow to ACK websocket messages, which prevents Tornado infinitely backlogging outgoing messages until it runs out of memory.  When using `workerSocket` the backlogging is in communication between the `Worker` and the main thread.  A slow application can include anything rendering graphics not achieving animation deadlines.

It still decodes messages in the main thread, so it's still important to choose a performant compression method.  This was done to minimize the overhead of posting data from the `Worker` to the main thread. 
 CBOR is ideal because the buffer can be transferred from the `Worker` to the main thread without copying.  Other compressions post a single string which is known to be more performant than posting an object.

Decoding messages in the `Worker` might be a good iteration on this, I would want to prove that it can be done without increasing post overhead too much.  It did not seem possible to integrate without heavily refactoring the socket layer of ROSLIB.

Making it work in Node and setting it default fully realizes `workerSocket`.  This iteration should be useful for most users of ROSLIB for visualization.

viktorku:
Is there any specific flow control mechanism in the worker (or main) thread such that the client runs OOM instead of the server? Or is it implicit, due to the non-blocking worker thread ACK-ing incoming data to rosbridge at a much faster rate (immediately upon receive) than the main thread is able to process/decode? I assume the latter.

This seems to make sense, as either the worker or main client thread would crash. I would say not ideal flow control but better to crash the slow clients than the server. Also, does this 100% guarantee backpressure would not propagate to Tornado?

I think we can improve on this (in future) by dropping "stale" data from the worker by sampling or similar RxJS [throttling techniques](https://github.com/ReactiveX/RxJava/wiki/Backpressure#useful-operators-that-avoid-the-need-for-backpressure).

:+1:

mvollrath:
Now that the [server connects backpressure](https://github.com/RobotWebTools/rosbridge_suite/pull/385) designing the application to properly queue high bandwidth topics is effective for flow control, where rosbridge_library used to hand everything off to Tornado's invisible infinite queue.  It makes [single-item queueing most topics in ROS3D](https://github.com/RobotWebTools/ros3djs/pull/254) an effective mitigation when visualizing things like point clouds.

`workerSocket` may be somewhat redundant with the server-side changes.  I feel like it will make more sense when decoding happens in the Worker, to really leverage concurrency and reduce messaging overhead on the main thread.  Until then it's a way to protect the server when the main thread is saturated with work other than handling ROSLIB messages.

chapulina:
> Can you confirm it works with your usecase? Then, we can just proceed to merge.

@jubeira checked that this works well with RVizWeb: https://github.com/osrf/polymer-ros-websocket/pull/4#pullrequestreview-212880609

jihoonl:
Thanks for all nice discussion and information. It looks all good in codewise as well. Thanks @mvollrath again.

chapulina:
Hi @jihoonl, thanks for reviewing and merging! Are there any plans for a release anytime soon? Thanks!

PR #320, opened 2019-02-09
Removing build directory in develop branch.

jihoonl:
build directory has been making conflicts with a series of Pull requests. This PR is to remove build folder in the source to manage the code cleaner.


mvollrath:
Some downstream bower projects [import the build](https://github.com/osrf/polymer-ros-websocket/blob/polymer1/2/roslib-import.html).  More thoughts on what to do about it later.

Rayman:
Bower is deprecated so users should migrate away from it

jihoonl:
I will reopen this later when I can follow up.

Issue #321, opened 2019-02-22
removeListener only takes instances of Function


baptistemanson:
From time to time, we get those logs in prod:
```
api | /usr/src/api/node_modules/eventemitter2/lib/eventemitter2.js:541
api | throw new Error('removeListener only takes instances of Function');
api | ^
api |
api | Error: removeListener only takes instances of Function
api | at RosTCP.EventEmitter.off (/usr/src/api/node_modules/eventemitter2/lib/eventemitter2.js:541:13)
api | at Topic.unsubscribe (/usr/src/api/node_modules/roslib/src/core/Topic.js:128:14)
```

We call unsubscribe like this:
```js
topic.unsubscribe()
```
It seems to be a race condition.

biw:
This might be fixed by upgrading to [`1.0.1`](https://github.com/RobotWebTools/roslibjs/releases/tag/1.0.1)

biw:
This is due to calling `topic.unsubscribe()` before `topic.subscribe()`. Created a fix in https://github.com/RobotWebTools/roslibjs/pull/324

baptistemanson:
Your patch fixed the issue thanks!

PR #324, opened 2019-02-25
added check for reconnectFunc in unsubscribe/unadvertise functions

biw:
In the Topic file, there is currently a case in which the library will pass an undefined value to the underlying `EventEmitter`.

In the `Topic` constructor, `this.reconnectFunc` is defined in the `this.callForSubscribeAndAdvertise` callback function, which is called in `this.subscribe()`. In most cases this is not an issue, however, there is one case where this will throw an error. 

If `this.reconnect_on_close === true` and `this.unsubscribe` is called before `this.subscribe` (which may be something that `roslibjs` should throw an error message on, thoughts @mvollrath?), the underlying `EventEmitter` [will throw](https://github.com/EventEmitter2/EventEmitter2/blob/master/lib/eventemitter2.js#L592) (as mentioned in #321):

```js
throw new Error('removeListener only takes instances of Function');
```

since `this.reconnectFunc` is `undefined`.

I changed the code to check to make sure that `this.reconnectFunc` is not undefined, which I think is the easiest fix, but happy to add more (such as an error telling the user they must subscribe before unsubscribing) 😄.


mvollrath:
I don't see how it allows `unsubscribe()` before `subscribe()`.  It is checking for `this.subscribeId` [here](https://github.com/RobotWebTools/roslibjs/blob/14e6cd1fc00ed73ba259efb7d2edf7f3611d256a/src/core/Topic.js#L125) which isn't even set until `subscribe()` [here](https://github.com/RobotWebTools/roslibjs/blob/14e6cd1fc00ed73ba259efb7d2edf7f3611d256a/src/core/Topic.js#L97).  Fixing the intention of this logic should prevent the exception.

biw:
Ahhh, you're 100% right @mvollrath. I'm a little confused now... Any idea why it would be throwing an error?

mvollrath:
I would start by writing a test that reproduces the error.  Something like https://github.com/RobotWebTools/roslibjs/blob/9297a87548c9db81dae6f0913fc193fb833dac5c/test/examples/pubsub.example.js might be a good starting point.

biw:
I'm not able to reproduce the error with test, so I'm going to close this until I can reproduce it 😄

Thanks for your help @mvollrath 😄

PR #348, opened 2019-12-21
WS 3.3.1 > 7.2.1

### LINKED ISSUE #347: Update WS version
### Problem

I experienced a problem when connection to the rosbridge was lost, because the rosbridge node was stopped.

The first line of code I could track down is:
https://github.com/websockets/ws/blob/3.3.3/lib/WebSocket.js#L188
On this line the `ultron` socket is destroyed. Which calls the remove function of the `ultron` socket on
https://github.com/primus/ultron/blob/1.1.0/index.js#L129
In the remove function on https://github.com/primus/ultron/blob/1.1.0/index.js#L81

The `eventNames` of the implemented eventEmitter retrieved. The resulting eventNames are:
`["end", "finish", "_socketEnd", "__ob__", "drain", "close", "error", "data"]`.

On https://github.com/primus/ultron/blob/1.1.0/index.js#L96, the items of this list are passed to `ee.listeners` which is on https://github.com/nodejs/node/blob/v11.15.0/lib/events.js#L411.

In this function the actual event is retrieved from the eventName. For the eventName `__ob__`, this causes an error. As the event retrieved from `__ob__` is a observer. So the checks, both the one if `evlistener` is undefined and if `evlistener` is a function, result to false. So this causes `unwraplistener` to be called on https://github.com/nodejs/node/blob/v11.15.0/lib/events.js#L425, with an observer as argument. While the `unwraplistener` only accepts arrays. Therefore on https://github.com/nodejs/node/blob/v11.15.0/lib/events.js#L475 a TypeError is raised as the first element of the observer doesn't exist.

### Solution
As this is part of the WS library on version 3.3.3, which is released on dec 17th, 2017. I expect `WS` will not support any issues regarding this problem. WS is now as version 7.X. Also `WS` got rid of the `ultron` implementation from 4.1.0. Therefore I ask you to update your `WS` version.
I shortly tested my application while using `roslibjs` with WS@4.1.0. This solved the problem for me.
In my opinion it would be the best to find the newest version of `WS` which would work with `roslibjs`

Update: Also a short test with WS@7.2.1 solved the problem and didn't show any other issues.

MatthijsBurgh:
Fixes #347


MatthijsBurgh:
ping ;)

MatthijsBurgh:
Ping ;)

MatthijsBurgh:
@mvollrath thanks for updating with develop branch. Could you please review it?

MatthijsBurgh:
@mvollrath could you create new release after merging this PR?

MatthijsBurgh:
@mvollrath Anything blocking for merging this PR?

PR #351, opened 2020-01-04
Add cbor-raw compression

janpaul123:
And a `getTopicsAndRawTypes` service function.

This is a companion PR to https://github.com/RobotWebTools/rosbridge_suite/pull/452. Please see that PR for more details and motivation.


src/core/Ros.js
```diff
@@ -624,5 +625,37 @@ Ros.prototype.decodeTypeDefs = function(defs) {
   return decodeTypeDefsRec(defs[0], defs);
 };
 
+/**
+ * Retrieves list of topics and their associated type definitions.
+ *
+ * @param callback function with params:
+ *   * topics - Array of topic names
+ *   * types - Array of message type names
+ *   * typedefs_full_text - Array of full definitions of message types, similar to `gendeps --cat`
```
mvollrath: Would you mind documenting `failedCallback` here?

janpaul123: Yeah this was missing throughout this file. I added a commit that adds this all over the file (based on https://github.com/RobotWebTools/roslibjs/blob/c10d9eca43ad51cc89b335d00e81f647cbcb77d3/src/core/Service.js#L36) and also fixes some other documentation strings in minor ways here and there.

PR #357, opened 2020-03-20
Add fail callback and missing docstrings to params

epaezrubio:


epaezrubio:
I'm looking forward to some feedback about this PR 😃

mvollrath:
Looks good, tests are unhappy though.

epaezrubio:
I'm unsure how to run the tests. If I have done it correctly, the tests pass in my machine. Moreover, how are the failed fibonacci tests related to these changes?

mvollrath:
This seems relevant: https://github.com/RobotWebTools/rosbridge_suite/issues/528

MatthijsBurgh:
@EzraBrooks @sea-bass I think this PR is worth reviewing.

sea-bass (commented):
I think this is a great addition.

Would you be able to modify the `simple.html` example to show this functionality?

EzraBrooks (approved):
This is a simple enough change, since it's just surfacing the underlying service callbacks, that I don't feel it needs additional testing or examples (given that the JSDoc comments indicate the usage). It's also an important addition - I myself have an application where I need to handle Param set failures in which I currently cannot without this change.

Thanks!

MatthijsBurgh:
@EzraBrooks is this a breaking change? I need to know whether to include this change in the final v1 release.

EzraBrooks:
This is non-breaking, it only added an optional argument.

PR #359, opened 2020-04-09
fix robot material color

ahoarau:
This happens when the name of the material is empty.
Everything appears gray. This simple check fixes it !

### DIFF
--- a/src/urdf/UrdfModel.js
+++ b/src/urdf/UrdfModel.js
@@ -72,10 +72,14 @@ function UrdfModel(options) {
         {
           var mat = link.visuals[j].material; 
           if ( mat !== null ) {
-            if (this.materials[mat.name] !== void 0) {
-              link.visuals[j].material = this.materials[mat.name];
-            } else {
-              this.materials[mat.name] = mat;
+            // Check if color is already present. If not check on the materials list.
+            if( mat.color === null )
+            {
+              if (this.materials[mat.name] !== void 0) {
+                link.visuals[j].material = this.materials[mat.name];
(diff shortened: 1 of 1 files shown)


ahoarau:
Any maintainers out there ?

mvollrath:
Looks like nobody has changed this file in 5 years, so you are the expert now.  Is that ok?

ahoarau:
There's a few issues on robot appearing all grey.
This happens (still !) if the material name is empty ( == "", rgb still set). It cannot get the rgb because it is considered as "in the materials list"

```

**Comment by mvollrath** [comment]
> This happens when the name of the material is empty.
> Everything appears gray.

I think I understand this.  When there is no material specified the model is gray.

> This simple check fixes it !

How does this check fix it?  Does it make the model some color other than gray?

> There's a few issues on robot appearing all grey.
> This happens (still !) if the material name is empty ( == "", rgb still set).

At a glance it looks like materials with colors set will not go through either of the old code paths after this change.  Can you explain how this change fixes the issue?

**Comment by ahoarau** [comment]
```xml
<link name="link_1">
      <material name="">
        <color rgba="0 1 0 1" />
      </material>
  </link>
<link name="link_2">
      <material name="">
        <color rgba="1 0 0 1" />
      </material>
  </link>
```

The material set in the link is overritten here: 
image

It is considering materials with the same empty names the same, even though the color values are different. This behavior is different from rviz.

Rayman:
Should be fixed by https://github.com/RobotWebTools/roslibjs/pull/374

PR #374, opened 2020-09-07
Support the case material's name is empty string

buster84:
Some robot's definition has empty string for material's name. And the empty string caused `this.materials` mess up. 
This fix the problem.

### DIFF
--- a/src/urdf/UrdfModel.js
+++ b/src/urdf/UrdfModel.js
@@ -71,7 +71,7 @@ function UrdfModel(options) {
         for( var j=0; j<link.visuals.length; j++ )
         {
           var mat = link.visuals[j].material; 
-          if ( mat !== null ) {
+          if ( mat !== null && mat.name ) {
             if (this.materials[mat.name] !== void 0) {
               link.visuals[j].material = this.materials[mat.name];
             } else {


Rayman (approved):

gavanderhoorn:
Related: ros-industrial/fanuc#297.

Rayman:
So are materials defined to be optional in urdf?

gavanderhoorn:
> So are materials defined to be optional in urdf?

The technical answer to that would be "yes", as you're not required to use any `material` tags in a URDF.

But I believe you're asking whether the `name` attribute on `material` elements is optional. If that's the case, then: no, it's not optional.

What we're doing in `ros-industrial/fanuc` though is using empty names for locally defined `material`s. That would be something else.

buster84:
I think what @gavanderhoorn said is correct. 

The official c++ implementation also has the special handling for empty string. See [here](https://github.com/ros/urdfdom/blob/master/urdf_parser/src/model.cpp#L64-L67)

In my opinion, I don't think there is a technical reason why `name` attribute on `material` elements is required because empty string is allowed. But it has been there for long time and it is hard to change the definition I guess.

buster84:
@Rayman I think CI is failing because of this issue https://github.com/ros/genpy/issues/132 . 
We need to add `apt update && apt dist-upgrade` in dockerfile. Is that okay to add in this PR or should I create another PR?

Rayman:
If I read ros/genpy#132 it makes sense to add `apt-get upgrade` to the dockerfile. You could try to fix the CI for this PR

I'm not a maintainer though, I just have a personal interest in this package.

buster84:
> If I read [ros/genpy#132](https://github.com/ros/genpy/issues/132) it makes sense to add `apt-get upgrade` to the dockerfile. You could try to fix the CI for this PR
> 
> I'm not a maintainer though, I just have a personal interest in this package.

@gavanderhoorn But you can merge this PR and release new version, can you? anyway, I made CI passed so please approve this or let me know if I need to do something else.

Rayman:
I can't, you would have to ask @mvollrath

Dockerfile
```diff
@@ -4,6 +4,9 @@ ENV ROS_DISTRO=kinetic
 # Dependencies for rosbridge
 RUN apt update && apt-get install -y xvfb firefox git wget ros-$ROS_DISTRO-rosbridge-server ros-$ROS_DISTRO-tf2-web-republisher ros-$ROS_DISTRO-common-tutorials ros-$ROS_DISTRO-rospy-tutorials ros-$ROS_DISTRO-actionlib-tutorials
 
+# Because of this issue https://github.com/ros/genpy/issues/132
+RUN apt update && apt-get upgrade -y
```
mvollrath, line 8: Doing this tends to cost a lot of time and space, and makes images less consistent.  The base image should be updated soon.  I appreciate that this is needed for the tests to pass, I will merge it without this change.

mvollrath (changes requested):

buster84:
@mvollrath thank you for the reviewing. I understood and took away the commit.

buster84:
BTW, https://github.com/RobotWebTools/roslibjs/pull/359 is also the same problem and I hope this fix solves the problem.

mvollrath (approved):

PR #375, opened 2020-09-17
Properly handle unsubcribe for TFClient

J-Rojas:


Rayman:
Can you explain what was wrong with the old implementation and how this PR improves it?

J-Rojas:
The old method never properly unsubscribes the processTFArray callback. It must be provided as an argument to the Topic.unsubscribe method to properly be unregistered.

mvollrath (approved):

PR #391, opened 2020-12-14
[temp](ci) install python-future

MatthijsBurgh:
python-future should be installed because rosbridge-library 0.11.11-0.11.12 are missing this dep. When 0.11.13 is available this manual install should be removed


MatthijsBurgh:
`0.11.13` has been released, so issue is resolved.

PR #392, opened 2021-01-08
Replace travis by GH actions

MatthijsBurgh:
Actions need to be enabled by maintainers

PR #412, opened 2021-04-04
Accept ros distro as argument in Dockerfile

MatthijsBurgh:


MatthijsBurgh:
@jihoonl In https://github.com/RobotWebTools/roslibjs/pull/392 we missed that everything was still running in kinetic. Despite setting the `ROS_DISTRO`. That should be fixed by this PR.

However we use `tf2_web_republisher` in our tests. Though it hasn't been released yet on noetic. We should either remove the noetic test or release it for noetic as well. Which will make other people happy too (https://github.com/RobotWebTools/tf2_web_republisher/issues/30).

jihoonl:
I will release tf2 web republisher soon.

jihoonl:
https://github.com/ros/rosdistro/pull/28945

Issue #322, opened 2019-02-25
What server to run for socket.io support


Rayman:
I'm trying to connect to rosbridge via socket.io as explained in the docs:
```js
new Ros( {..., transportLibrary: 'socket.io'} );
```
What server supports this? I'm now running `roslaunch rosbridge_server rosbridge_websocket.launch` but this only supports websockets I think (I'm getting CORS errors).

moth-man:
Did you ever figure out the problem? Getting the same error

Rayman:
Unfortunately no, I'm now running a normal websocket with some reconnect logic.

VadimCiv:
Same problem here. Trying to use Ros with  transportLibrary: 'socket.io' option  but still get error "ReferenceError: io is not defined"
Please explain what I doing wrong?

package.json "socket.io": "^4.1.3",

```
var ROSLIB = require('roslib');
var io = require('socket.io');

var ros = new ROSLIB.Ros({
  transportLibrary: 'socket.io',
});

export const connectToRos = (ip) => {
  ros.connect(`ws://${ip}:9090`); // *** error here ***
};

```
Ros.js
```

Ros.prototype.connect = function(url) {
  if (this.transportLibrary === 'socket.io') {
    this.socket = assign(io(url, {'force new connection': true}), socketAdapter(this));
...
}
```

ReferenceError: io is not defined
    at Ros.push../node_modules/roslib/src/core/Ros.js.Ros.connect (Ros.js:66)

VadimCiv:
Same here ReferenceError: io is not defined
Anybody can explane how use transportLibrary: 'socket.io'?

rohanshukla9:
Even I am getting the same error, any workarounds?

EzraBrooks:
socket.io support has been removed in 2.x to decrease maintenance surface.

PR #429, opened 2021-04-21
Add missing socket.io to Ros.js

MatthijsBurgh:


Rayman:
Does the socket.io layer even work? I've asked the question before but I didn't get an answer: https://github.com/RobotWebTools/roslibjs/issues/322

MatthijsBurgh:
@Rayman I am not sure it works. It isn't tested. 

But at least I know `io` wasn't imported.

J-Rojas:
I believe the reason why this is left out is because socketio is non-essential and an optional integration whereas the other dependencies are critical to the function of roslib. Socketio is a relatively large library (60kb minified) larger than roslib is currently. So I would keep it as a peer dependency and not require it here.

MatthijsBurgh:
I am totally fine by making it a peer dependency, but right now it is marked as a normal dependency.

Also right now `io` is just not available inside `ros.js`

francoissunatori (approved):

francoissunatori:
Why wasn't this already merged?
[`socket.io` is already in `package.json`](https://github.com/RobotWebTools/roslibjs/blob/develop/package.json#L41)

MatthijsBurgh:
@francoissunatori see https://github.com/RobotWebTools/roslibjs/pull/429#issuecomment-825789600

The question remains if it works and whether it should be included in the compiled libs. If not included a user can always import it him/herself.

PR #437, opened 2021-05-10
change dependbot schedule to montly

jihoonl:
Until https://github.com/dependabot/dependabot-core/issues/1190 gets resolved. It seems to create extreme noise at the moment.

@MatthijsBurgh  how do you think? Or I wonder if we can automatically merge these PRs. It seems too noisy as well as makes clicking burden.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,4 +3,4 @@ updates:
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"


MatthijsBurgh:
I think weekly is OK for `roslibjs`. Though `ros3djs` is creating too much noise, because of the examples. So lets see if we can ingore the examples folders in `ros3djs`.

MatthijsBurgh (changes requested):
Favoring #438 over this one

PR #438, opened 2021-05-10
Widen package requirements

MatthijsBurgh:
As suggested in https://github.com/RobotWebTools/roslibjs/pull/437#issuecomment-836589129

Closes: https://github.com/RobotWebTools/roslibjs/pull/437


package.json
```diff
@@ -14,32 +14,32 @@
...
   },
   "dependencies": {
     "cbor-js": "^0.1.0",
     "eventemitter2": "^6.4.0",
-    "object-assign": "^4.0.1",
-    "pngparse": "^2.0.1",
-    "socket.io": "4.0.2",
+    "object-assign": "^4.0.0",
+    "pngparse": "^2.0.0",
+    "socket.io": "4.0.0",
```
jihoonl, line 40: I think it should be `^4.0.0`? Or should this use the exact version?

MatthijsBurgh: This is wrong, fixed it

jihoonl (approved):

jihoonl:
👍

PR #465, opened 2021-08-09
Add a qos_durability option to subscribe

MrBlenny:
I needed the ability to subscribe to either `transient_local` or `volatile` messages when using the `rosbridge_server`.

This PR adds a `qos_durability` flag to subcription requests which is then acted upon in the `rosbridge_server`. The server-side logic is in [another pull request](https://github.com/RobotWebTools/rosbridge_suite/pull/582)

I am very new to ROS so I'm very keen for some comments here. I did this very much from a ROS2 perspective - not sure about ROS1 or how this fits in with the [rosbridge_protocol spec](https://github.com/RobotWebTools/rosbridge_suite/blob/develop/ROSBRIDGE_PROTOCOL.md)


MatthijsBurgh:
I don't use ROS2 yet. So I can't review this one. Maybe someone from @RobotWebTools/ros2-0 has the knowledge to do so.

PR #469, opened 2021-08-24
URDF Parser gives NaN value Poses

Nickick-ICRS:
The URDF parser gives NaNs in Poses if the URDF doesn't have exactly 1 space in between each number.

I've made a change to use a regex in the split function, so as to separate numbers with any amount of white-space in between
I've also updated the URDF unit tests to confirm that the change works.

### DIFF
--- a/src/urdf/UrdfJoint.js
+++ b/src/urdf/UrdfJoint.js
@@ -44,7 +44,7 @@ function UrdfJoint(options) {
     var xyz = origins[0].getAttribute('xyz');
     var position = new Vector3();
     if (xyz) {
-      xyz = xyz.split(' ');
+      xyz = xyz.split(/\s+/);
       position = new Vector3({
         x : parseFloat(xyz[0]),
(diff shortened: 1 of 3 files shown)


Nickick-ICRS:
You can see the issue I'm fixing by running the (new) tests on commit 368bc473cb23f06130aef60442f4f875f267a759

MatthijsBurgh:
Although I can't find the strict spec of URDF(So let me know if you can). I think it the correct way to have vectors a single space separated string of values.

I am pro for making robust code, but I don't think we should support more than necessary. So right now I am more in favor of raising an error than supporting any whitespace.

P.S. https://github.com/ros/urdfdom_headers/blob/8ec38993c21c6913602b9519ca07e4e018af88b5/urdf_model/include/urdf_model/pose.h#L66 Is the closest I have come to URDF spec.

MatthijsBurgh:
@Nickick-ICRS any comments? Else I will close this PR.

PR #475, opened 2021-09-12
[WIP] convert to ES Module format

trusktr:
**Public Changes**

None

**Description**

Converted to ES Modules (still ES5 classes though)

I'm doing this so that ros3d will be as close to fully ESM-compliant as possible (related PR in ros3d to convert to `class` syntax: https://github.com/RobotWebTools/ros3djs/pull/438).

We may be able to complete that ros3d PR without this one (the source won't be as ESM-compatible, but the build output will be). We can complete both of these PRs without being 100% native ESM-compatible, but at least a lot closer (replacing all dependencies with ESM-ready versions would be the biggest step needed).

This is part of my longer term modernizing goal which includes the following in no particular order:

- convert to TypeScript (need to convert roslib to classes first)
- optionally compile to WebAssembly (with AssemblyScript, TS to Wasm compiler)
- update Three.js to latest
- replace unnecessarily big dependencies like EventEmitter2 (1600 lines of code for something that can be done in 100).

## TODO

- [ ] Run tests
- [ ] Manually test Node.js usage
- [ ] Manually test browser usage
- [ ] Add shim module for `bson` in browsers
- [ ] Restore WorkerSocket. I disabled some shims to get it to compile for browsers first. These need to be re-shimmed in the Rollup build.


src/RosLib.js
```diff
@@ -3,26 +3,10 @@
...
-
-assign(ROSLIB, require('./math'));
-
-assign(ROSLIB, require('./tf'));
-
-assign(ROSLIB, require('./urdf'));
+export const REVISION = '1.1.0';
 
-module.exports = ROSLIB;
+export * from './core/index.js';
```
Rayman, line 8: Do you need to hardcode `.js` for all imports? Is that mandatory?

Rayman (commented):
I like these changes. I've been thinking about starting this myself, but never got the time to do it

src/RosLib.js
trusktr: It is mandatory by default in Node.js ESM and browser ESM.

trusktr:
I just pushed an update that completes initial ESM format, and builds the output in three formats: ESM, CJS, and UMD. The `./build/roslib.js` and `./build/roslib.mni.js` file changed to be UMD format, but the end result is the same for those two files thus they are backward compatible. 

The new `.cjs.js` file is there in case anyone needs it for some reason (maybe they have an older version of Node.js that supports only CommonJS).

The new `.esm.js` file is good for cases when someone needs to import the whole library using native ESM in a browser, where the browser (or more specifically the web app's server) will not know how to lookup Node-style import specifiers like `import foo from 'foo'`. The browser will try to fetch `current-file-folder/foo` which will not be found. So the single ESM bundle solves this case. Newer versions of Node.js should be able to import directly from the source files

I haven't done any testing yet (adde TODOs to the OP).

src/core/Ros.js
```diff
@@ -3,15 +3,17 @@
  * @author Brandon Alexander - baalexander@gmail.com
  */
 
-var WebSocket = require('ws');
-var WorkerSocket = require('../util/workerSocket');
-var socketAdapter = require('./SocketAdapter.js');
+import WebSocket from 'ws';
+// import {WorkerSocket} from '../util/workerSocket.js'; // TODO
```
trusktr, line 7: - [ ] TODO I disabled some shims to get it to compile for browsers first. These need to be re-shimmed using Rollup.

Also another question I have is, is WorkerSocket actually useful? Has there been performance gains from this? I'm skeptical because fetching from network is a non-blocking operation. Does the data handling (from the browser engine, before finally passing that data to JS) have a big cost that the worker is putting on a separate thread?

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "The standard ROS Javascript Library",
   "version": "1.1.0",
   "license": "BSD-2-Clause",
+  "type": "module",
```
trusktr, line 7: This is the main breaking change (major version bump)

MatthijsBurgh:
@trusktr already thanks for your efforts.

Could you please enable GH actions for your repo. At the moment we only run on pushes, no PRs. But I think we should revert that.

trusktr:
On my end I see it is waiting for actions to run:

Screenshot from 2021-09-14 13-21-23

trusktr:
Alriiiiiight, tests are running (https://github.com/trusktr/roslibjs/runs/3606731719?check_suite_focus=true). Tomorrow is update-from-jshint-to-eslint day! It'll be nice to have the same tooling setup in both roslib and ros3d.

src/core/Ros.js
Rayman: I think (but I'm not sure) that the CPU heavy work is decompressing the messages. Suppose you are sending a pointcloud with png based message compression. This work will then happen in a worker.

trusktr:
eslint is working, build is working (but the output not tested yet).

trusktr:
I think the idea of being compatible with native browser ESM will have to wait, but at least the code will be essentially ES format with the import/export syntax, while the output `build/roslib.esm.js` file will still satisfy the native-browser-ESM case. There is a bit of extra work to do it: namely converting roslib's dependencies to ESM too.

Rayman:
I would advise to keep the changes focussed on one topic as this makes it easier to review. Please create additional PRs for aditional topics (such as changing and fixing linting)

src/core/Ros.js
trusktr: That makes sense, but unless I overlooked it, it seemed to me that only the WebSocket network request runs in the worker, and passes the data back to the main thread to process the PNG. I don't see that `decompressPng` file being imported in the worker, I just see it `postMessage`ing data.

trusktr:
> I would advise to keep the changes focussed on one topic as this makes it easier to review. Please create additional PRs for aditional topics (such as changing and fixing linting)

That's what I thought initially, but the syntax broke `jshint`, so I figured that rather than try to update a dated tool, I'd just put the new tool in. After putting ESLint into place, there were very minimal code changes (just some missing semicolons, and a few unedfined vars, but all else stayed the same due to the lax rules that came from the ros3d repo).

I also thought about code formatting (f.e. with Prettier), but then that would cause changes _everywhere_ and thus any change in develop would conflict with this PR, so I definitely held off on that idea.

trusktr:
I have mocha tests passing locally on my end now, just need a little more work on the karma side (this time shimming import as opposed to shimming require).

src/core/Message.js
```diff
@@ -3,16 +3,14 @@
...
+import assign from 'object-assign';
 
 /**
  * Message objects are used for publishing and subscribing to and from topics.
  *
  * @constructor
  * @param values - object matching the fields defined in the .msg definition file
  */
-function Message(values) {
+export function Message(values) {
```
trusktr, line 14: Even with the ESLint update, the main source code changes are the import statements, not much else changed.

package.json
```diff
@@ -4,40 +4,50 @@
...
+    "rollup-plugin-terser": "^7.0.2",
     "time-grunt": "^2.0.0"
   },
   "dependencies": {
     "cbor-js": "^0.1.0",
     "eventemitter2": "^6.4.0",
     "object-assign": "^4.0.0",
     "pngparse": "^2.0.0",
-    "socket.io": "^4.0.0",
+    "socket.io": "trusktr/socket.io#bf4bd150cff3c043c1f60ddf347e1b3ef0f1339a",
```
trusktr, line 50: This is not idea (see https://github.com/socketio/socket.io/pull/4100). The alternative would be to update the `socket.io` import sites to import from a local file that shims the missing default function (`(srv, opts) => new Server(srv, opts);`) in Node.js, but in the browser just uses the `io` global.

src/util/shim/socket.io.js
```diff
@@ -0,0 +1,8 @@
+export default thro
+export const Server = thro
+export const Namespace = thro
+export const Socket = thro
+
+function thro() {
+    throw new Error('transportLibrary "socket.io" is not supported in browsers. Set the `transportLibrary` option to "socket.io" only if you are use roslib in an environment like Node.js or Electron.')
+}
```
trusktr, line 8: - [ ] TODO this file is totally wrong, I didn't realize socket.io works in a browser too. Update this so that it exports the global stuff from a global `io` variable from the global-script-tag version of socket.io.

trusktr:
Alrighty, the base `npm test` (mocha and karma) tests pass. Now I need to do similar for the test-examples and test-tcp (and maybe test-workersocket if that actually matters, please confirm if it is actually running PNG decompression in a worker, or just merely making network requests and sending data back to main thread).

test/karma.conf.cjs
```diff
@@ -4,19 +4,27 @@ module.exports = function(config) {
...
+      {pattern: 'src/util/shim/cbor.js', serve: true, included: false},
+      {pattern: 'src/util/cborTypedArrayTags.js', serve: true, included: false},
+
+      // This file maps some import specifiers to our module shims.
+      'test/util/write-importmap.js',
+
+      // Installs cbor-js in a global CBOR variable so that our shim/cbor.js shim picks it up.
+      'node_modules/cbor-js/cbor.js',
+      
+      {pattern: 'test/*.test.js', type: 'module'},
```
trusktr, line 27: `type: 'module'` makes this file operate as native browser `<script type="module"` , and therefore the passing tests are exercising the new `roslib.esm.js` output file in browser ESM.

Gruntfile.cjs
```diff
@@ -10,34 +12,39 @@ module.exports = function(grunt) {
...
-      files: [
-        './Gruntfile.js',
-        './src/**/*.js'
-      ]
     },
     karma: {
       options: {
         singleRun: true,
-        browsers: process.env.CI ? ['FirefoxHeadless'] : ['Firefox']
+        browsers: process.env.CI ? ['ChromeHeadless'] : ['Chrome']
```
trusktr, line 38: I switched to Chrome because I was having issues getting the `importmap` polyfill working in Firefox:

- https://github.com/guybedford/es-module-shims/issues/186
- https://github.com/guybedford/es-module-shims/issues/187

I decided to just get this all working natively first, then we can polyfill `importmap` in Firefox later if we still want to. But MS Edge is now built on Chromium, so at least the tests currently cover a large user base.

trusktr:
I noticed a lot of time is spent building the `roslibjsdocker` container each time just to run the tests. We can publish this container image on docker and just re-use each time. That'll speed CI up quite a bit.

MatthijsBurgh:
@trusktr what is the status of this? I really appreciate your work and would like to merge this when ready.

PR #486, opened 2021-10-22
Enable to unsubscribe tf in the callback

buster84:
**Public API Changes**

None

**Description**

The tfclient had a problem to unsubscribe in the callback because `splice` doesn't work well in the iterating. 
This PR fixed the problem to copy callback function array first before iterating so that the copied array doesn't get the affection from the unsubscribe.


MatthijsBurgh:
@EzraBrooks @sea-bass I don't know whether it is really possible listeners will be remove, while iterating over the callbacks. So I don't know whether this PR is relevant. Please determine whether it is relevant. In that case we should check whether the suggested changes fix the issue. Or we should come up with a different fix.

PR #489, opened 2021-10-27
Prevent race condition in TFClient

### LINKED ISSUE #488: Subscription leak when disposing TFClient while service request is ongoing
**Description**
If the TFClient is disposed while a service request to update the TF-republishing is ongoing. Then when the service call callback is called, it will resubscribe to the republished topic. Noone is using the TFClient anymore but it will stay subscribed to the republish topic, preventing the tf2_web_republisher from shutting down that republisher.

**Steps To Reproduce**
```
// We want to use the service interface
rosClient.groovyCompatibility = false;

const tfClient = new TFClient({ros: rosClient});
tfClient.subscribe('whatever', () => {})

// Only happens if the dispose call is called during the service call. So this timeout needs luck to hit the correct time. A slow backend is preferable to reproduce this issue.
setTimeout(() => { tfClient.dispose() }, 50)
```

**Expected Behavior**
No subscriptions to the republished topic

**Actual Behavior**
The TFClient subscribes to the topic despite already being disposed.

jorgenfb:
**Public API Changes**
None

**Description**
If a TFClient is disposed while a service call to update the goal is ongoing it will subscribe to the republished topic even if already disposed. To prevent this we mark the TFClient as disposed when it is disposed, and check for this in the service callback.

Fixes #488

### DIFF
--- a/src/tf/TFClient.js
+++ b/src/tf/TFClient.js
@@ -52,6 +52,7 @@ function TFClient(options) {
   this.frameInfos = {};
   this.republisherUpdateRequested = false;
   this._subscribeCB = null;
+  this._isDisposed = false;
 
   // Create an Action client
   this.actionClient = new ActionClient({
@@ -144,6 +145,12 @@ TFClient.prototype.updateGoal = function() {
  * @param response the service response containing the topic name
  */
 TFClient.prototype.processResponse = function(response) {
+  // Do not setup a topic subscription if already disposed. Prevents a race condition where
+  // The dispose() function is called before the service call receives a response.
+  if (this._isDisposed) {
+    return;
+  }
+
   // if we subscribed to a topic before, unsubscribe so
   // the republisher stops publishing it
   if (this.currentTopic) {
@@ -216,6 +223,7 @@ TFClient.prototype.unsubscribe = function(frameID, callback) {
  * Unsubscribe and unadvertise all topics associated with this TFClient.
  */
 TFClient.prototype.dispose = function() {
+  this._isDisposed = true;
   this.actionClient.dispose();
   if (this.currentTopic) {
     this.currentTopic.unsubscribe(this._subscribeCB);
--- a/test/tfclient.test.js
+++ b/test/tfclient.test.js
@@ -1,0 +1,30 @@
+var expect = require('chai').expect;
+var ROSLIB = require('..');
+
+describe('TFClient', function() {
+
+  describe('dispose', function() {
+    
+    it('should not subscribe to republished topic if already disposed', function() {
+      // This test makes sure we do not subscribe to the republished topic if the 
+      // tf client has already been disposed when we get the response (from the setup request)
+      // from the server.
+
+      var dummyROS = {
+        idCounter: 0,
+        on: () => {},
+        off: () => {},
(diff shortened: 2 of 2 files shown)


MatthijsBurgh:
@jorgenfb, could we create a test, which fails in the old situation, but works with the fix?

jorgenfb:
I have added a tests that fails without my fix and also rebased on develop.

MatthijsBurgh:
The `karma:test` step fails locally on my computer. While it always showed the same results as CI. So I can't accept this change for now. The test needs to be more robust.
```
Running "karma:test" (karma) task
07 01 2022 08:48:19.965:INFO [karma-server]: Karma v6.3.9 server started at http://localhost:9876/
07 01 2022 08:48:19.966:INFO [launcher]: Launching browsers Firefox with concurrency unlimited
07 01 2022 08:48:19.969:INFO [launcher]: Starting browser Firefox
07 01 2022 08:48:23.434:INFO [Firefox 95.0 (Ubuntu 0.0.0)]: Connected on socket 9K9kfopYzzASTtETAAAB with id 60949544
WARN: 'CBOR 64-bit integer array values may lose precision. No further warnings.'
ERROR: '(node) warning: possible EventEmitter memory leak detected. 11 listeners added. Use emitter.setMaxListeners() to increase limit.'
Firefox 95.0 (Ubuntu 0.0.0) TFClient dispose should not subscribe to republished topic if already disposed FAILED
	expected { Object (ros, name, ...) } to be false
	AssertionError@/home/matthijs/src/roslibjs/node_modules/chai/chai.js:9574:13
	[3]</module.exports/Assertion.prototype.assert@/home/matthijs/src/roslibjs/node_modules/chai/chai.js:250:13
	[5]</module.exports/<@/home/matthijs/src/roslibjs/node_modules/chai/chai.js:1072:10
	propertyGetter@/home/matthijs/src/roslibjs/node_modules/chai/chai.js:7959:29
	proxyGetter@/home/matthijs/src/roslibjs/node_modules/chai/chai.js:9372:22
	@tfclient.test.js:26:13
	
Firefox 95.0 (Ubuntu 0.0.0): Executed 25 of 25 (1 FAILED) (0.063 secs / 0.022 secs)
TOTAL: 1 FAILED, 24 SUCCESS
Warning: Task "karma:test" failed. Use --force to continue.

Aborted due to warnings.
```

MatthijsBurgh:
@jorgenfb I didn't compile yet. Which caused the `karma` test to fail. Please restore the PR. Sorry for the confusion.

jorgenfb:
Hi Matthijs! Can you please check locally again. I don't see why this test would be flaky, I have gone through the logic, and it is pretty simple. I just reverted my fix, without removing the test, and the CI clearly fails now. 

I have also tried running this on my machine:
```sh
docker build --build-arg ROS_DISTRO=melodic --build-arg NODE_VERSION=14 -t roslibjsdocker .
docker run -v $(pwd):/root/roslibjs --rm roslibjsdocker bash -i -c 'bash /root/roslibjs/test/build.bash'
```
It works

jorgenfb:
Good! I will remove the last commit again. No problem, these things happen

MatthijsBurgh (approved):

PR #508, opened 2022-01-10
Ros logging

### LINKED ISSUE #352: roslibjs logging functions missing?
My team was looking into how to use ROS logging (loginfo, logerr, logwarn) in roslibjs for our project but it doesn't look like this functionality exists right now. If it does exist, how can I use it? Otherwise, I think it would be good to have.

joshua-glazer:
**Public API Changes**

Implements Javascript logging functionality similar to ROS logging functions `logwarn`, `logerr`, etc. 

**Description**

Created a JS file implementing functions which log to the console and publish to `rosout` based on the ROS logging level. Created an example file to demonstrate how the functions are used.

Resolves #352

### DIFF
--- a/examples/Logging.html
+++ b/examples/Logging.html
@@ -1,0 +1,77 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8" />
+<script src="https://static.robotwebtools.org/EventEmitter2/current/eventemitter2.min.js"></script>
+<script src="../build/roslib.js"></script>
+
+<script>
+  // Connecting to ROS
+  // -----------------
+  var ros = new ROSLIB.Ros();
+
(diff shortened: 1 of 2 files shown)


MatthijsBurgh:
@sea-bass @EzraBrooks What do you think of this PR?

sea-bass:
> @sea-bass @EzraBrooks What do you think of this PR?

I would probably prefer to have this be part of the protocol and have the Python backend do actual logging calls through `rclpy`.

I say this because publishing to `/rosout` is a limited subset of logging that notably does not make it into a log file -- see, e.g., https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Logging.html#overview

MatthijsBurgh:
I agree with the vision of @sea-bass @EzraBrooks therefore I am closing this PR.

PR #556, opened 2022-07-22
Add channel omission flags to ActionClient's JSDoc param tag

jamestiotio:
**Public API Changes**
This PR makes the following channel omission flags for `ActionClient` explicitly visible to the user via JSDoc's `@param` tag:
- `omitFeedback`
- `omitStatus`
- `omitResult`

**Description**
This change is in line with PR #202. This PR explicitly allows the aforementioned channel omission flags to be passed in as part of the `options` parameter of the constructor function of `ActionClient`.


MatthijsBurgh (approved):

MatthijsBurgh:
Thanks.

I think the GitHub hosted docs will only be updated on a new release. So you will have to wait for it to be available online. But you can generate the docs yourself.

jamestiotio:
> I think the GitHub hosted docs will only be updated on a new release. So you will have to wait for it to be available online. But you can generate the docs yourself.

No problem, thanks!

PR #597, opened 2023-02-20
React example

MatthijsBurgh:
superseeds #587


MatthijsBurgh:
@snorresovold things I changed, which are good to take into account for future examples:
- Use geometry_msgs/Pose2D instead of a custom msg
- Be consistent with linting
- Remove all unnecessary dependencies, code, file, etc.

MatthijsBurgh:
Preferably examples are tested in CI. But that requires to install and build it before it can be tested. Therefore not doing it.

PR #605, opened 2023-04-17
prevent calling on undefined

chengzegang:
**Description**
I keep getting this error:
```js
caught TypeError: Cannot read properties of undefined (reading 'encoder')
    at Ros.callOnConnection [as callForSubscribeAndAdvertise] (Ros.js:149:1)
    at Topic.advertise (Topic.js:150:1)
    at Topic.publish (Topic.js:194:1)
    at publishTopic (ros_node.ts:63:1)
 ```
So I made a small patch to do an undefined check:

```js
Ros.prototype.callOnConnection = function (message) {

  if (this && this.transportOptions && this.transportOptions.encoder) {
    this.transportOptions.encoder(message, this._sendFunc);
  } else if (this && this._sendFunc)  {
    this._sendFunc(JSON.stringify(message));
  }
};
```
And then everything works well.
This might be due to the async calls, so "this" is not ready.

### DIFF
--- a/src/core/Ros.js
+++ b/src/core/Ros.js
@@ -38,7 +38,7 @@ function Ros(options) {
   this.isConnected = false;
   this.transportLibrary = options.transportLibrary || 'websocket';
   this.transportOptions = options.transportOptions || {};
-  this._sendFunc = function(msg) { that.sendEncodedMessage(msg); };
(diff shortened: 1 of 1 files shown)


MatthijsBurgh (changes requested):
Revert all whitespace changes. After reverting it, I will review the actual changes.

PR #636, opened 2023-10-05
use default function parameters

m-tartari:
**Public API Changes**
None

**Description**
Use [default function parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) instead of reinitializing variables at the start of the function.


MatthijsBurgh:
This is an ES6 feature, so it doesn't work.

Issue #463, opened 2021-08-01
Adjust function Ros.prototype.decodeTypeDefs 


peach96:
I wanted to use the function **Ros.prototype.decodeTypeDefs** in my project to decode the results I got from the functions **Ros.prototype.getMessageDetails** and **Ros.prototype.getServiceRequestDetails**, but it only worked for simple message/request types.

Therefore I adjusted the existing function, so that it would also work for complex types, for example 'geometry_msgs/twist'. 
I thought I'd post my solution here, if anyone has the same problem:

```
 decodeTypeDefs(defs) {
        var that = this;

        // calls itself recursively to resolve type definition using hints.
        var decodeTypeDefsRec = function (theType, hints) {
            var typeDefDict = {};
            for (var i = 0; i < theType.fieldnames.length; i++) {
                var arrayLen = theType.fieldarraylen[i];
                var fieldName = theType.fieldnames[i];
                var fieldType = theType.fieldtypes[i];

                // check if fieldType can be found in fieldnames 
                var complexFieldType = (defs.findIndex(def => def.type === fieldType) !== -1) ? true : false;

                //if fieldType complex --> then change fieldarraylen to 0
                arrayLen = complexFieldType ? 0 : arrayLen;

                if (!complexFieldType) { // check if fieldType is complex
                    if (arrayLen === -1) {
                        typeDefDict[fieldName] = fieldType;
                    } else {
                        typeDefDict[fieldName] = [fieldType];
                    }
                } else {
                    // lookup the name
                    var sub = false;
                    for (var j = 0; j < hints.length; j++) {
                        if (hints[j].type.toString() === fieldType.toString()) {
                            sub = hints[j];
                            break;
                        }
                    }
                    if (sub) {
                        console.log(true);
                        var subResult = decodeTypeDefsRec(sub, hints);
                        if (arrayLen === -1) {
                        } else {
                            typeDefDict[fieldName] = [subResult];
                        }
                    } else {
                        that.emit('error', 'Cannot find ' + fieldType + ' in decodeTypeDefs');
                    }
                }
            }
            return typeDefDict;
        };
        return decodeTypeDefsRec(defs[0], defs);
    } 
```

EzraBrooks:
This appears to have been fixed in #638.

PR #638, opened 2023-10-16
Ensure `fieldType` is a scalar if `arrayLen` is `-1`

scottbell:
**Public API Changes**

None

**Description**

Before this change, `typeDefDict[fieldName]` was always assigned as an array, irrespective of other variables’ states. With this new update, if the `arrayLen` is equal to `-1`, the `typeDefDict[fieldName]` is assigned the non-array value `fieldType`. If `arrayLen` doesn't equal `-1`, the assignment remains an array format, `[fieldType]`.

We used this to successfully [decode ROS2 nested types](https://github.com/RobotWebTools/rosbridge_suite/pull/883).

#463

### DIFF
--- a/src/core/Ros.js
+++ b/src/core/Ros.js
@@ -637,16 +637,11 @@ Ros.prototype.getMessageDetails = function(message, callback, failedCallback) {
   }
 };
 
-/**
- * Decode a typedef array into a dictionary like `rosmsg show foo/bar`.
- *
- * @param {Object[]} defs - Array of type_def dictionary.
- */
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
@scottbell I will take a look at this, but it could take some time.

MatthijsBurgh:
@sea-bass @EzraBrooks what do you think of this PR? I see you approved the related PR in `rosbridge_suite`.

sea-bass (approved):
This one LGTM -- we are using the author's related `rosbridge_suite` capabilities in our tool as well for fixed-size arrays, so we can vouch for this work.

PR #643, opened 2023-11-01
Don't check in built content

EzraBrooks:
**Public API Changes**

None

**Description**

Removes the built content from the repository to avoid accidental creation of giant, hard-to-resolve diffs in PRs.

WIP/TODO: how to handle examples? should these require building? can we inject a browser require polyfill (if we're not already) to make the raw src work fine?


MatthijsBurgh:
The only wat I accept such a change is, when we attach builds to the releases and/or release branches.

package.json
```diff
@@ -55,7 +55,7 @@
   "scripts": {
     "build": "grunt build",
     "doc": "grunt doc",
-    "test": "grunt test",
+    "test": "npm run build && grunt test",
```
MatthijsBurgh, line 58: Not accepting this change. When you make your case I could accept `grunt build && grunt test`

EzraBrooks: `npm run build` here guarantees that if a user changes the `build` script, the changes are reflected here. Is your concern that it's NPM-specific? It seemed from the README that this project is developed with NPM.

EzraBrooks:
> The only wat I accept such a change is, when we attach builds to the releases and/or release branches.

Can you explain to me why the built content is checked into the repo? This is extremely atypical in my experience. The JavaScript ecosystem has package managers that work, we don't need to ship tarballs.

EzraBrooks:
Superseded by #655

PR #645, opened 2023-11-06
Add ROS 2 action support

sea-bass:
**Public API Changes**
This PR adds new ROS 2 functionality for actions, without touching any of the existing ROS 1 APIs.

This, plus the fact that CI currently is not running on any ROS 2 distros, does beg the question on what we should do with ROS 2... but I'll raise a different issue for this.

**Description**
This PR builds on the ROSBridge protocol updates in https://github.com/RobotWebTools/rosbridge_suite/pull/886, to add official action support for ROS 2.

Notably, this doesn't rely on any hidden topics/services or extra logic on the `roslibjs` side. Instead, `rosbridge_suite` is directly using the `rlcpy.actions` module which is way cleaner in my opinion.

The examples have been updated to also work with ROS 2, though I haven't done anything with the unit tests / CI yet. I think the state of testing is such that we'd need a host of separate PRs to address this. including actually running CI on ROS 2.

**Testing Instructions**
* Set up this package (`npm install` / `npm run build`)
* Open the `examples/ros2_action_client.html` and `examples/ros2_action_server.html` examples in your web browser and try them out, following the instructions
* Note you might need this `demos` repo installed as well: https://github.com/ros2/demos -- it has binaries, so it may already be on your system... but just in case.


src/actionlib/SimpleActionServer.js
EzraBrooks, line 1: unrelated formatting changes?

examples/simple.html
EzraBrooks, line 1: unrelated formatting changes?

src/actionlib/index.js
```diff
@@ -5,7 +5,6 @@ var action = module.exports = {
     ActionClient: require('./ActionClient'),
     ActionListener: require('./ActionListener'),
     Goal: require('./Goal'),
-    SimpleActionServer: require('./SimpleActionServer')
```
EzraBrooks, line 8: shouldn't this stay if we're keeping this for ros1 compat?

test/examples/fibonacci.example.js
EzraBrooks, line 1: unrelated formatting change?

EzraBrooks (commented):
looks good/sensical to me from a first pass, though I'll admit that my pre-`class`-keyword JavaScript is a bit rusty. Just some minor nits that I think all relate to the waffling over keeping ros1 actionlib support earlier on.

examples/simple.html
sea-bass: I think I want to keep these in, because it makes the comment style consistent within this same file ... but all the other formatting changes I will revert.

src/actionlib/index.js
sea-bass: oof yeah, missed this revert. Nice catch

src/core/Action.js
```diff
@@ -0,0 +1,162 @@
...
+  var actionGoalId = 'send_action_goal:' + this.name + ':' + (++this.ros.idCounter);
+
+  if (resultCallback || failedCallback) {
+    this.ros.on(actionGoalId, function(message) {
+      if (message.result !== undefined && message.result === false) {
+        if (typeof failedCallback === 'function') {
+          failedCallback(message.values);
+        }
+      } else if (message.op === 'action_feedback' && typeof feedbackCallback === 'function') {
+        feedbackCallback(new ActionResult(message.values));
```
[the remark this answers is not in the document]
sea-bass: thank you!

sea-bass:
I've tested this along with https://github.com/RobotWebTools/rosbridge_suite/pull/886 and https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67407 in our UI and all seems to work in order.

Inclined to merge this soon, as it looks ready to go now!

MatthijsBurgh (commented):
I have a few questions before merging this one:
- The ROS2 example, doesn't contain setting parameters. Is this still possible? In case it is, could we add it to the example?
- The changes in the rosbridge_server aren't released yet, correct? Therefore we can merge this one, but we shouldn't release it yet. Correct?
- This is not really a breaking change, but it only works with the ROS2 rosbridge_server correct? So therefore a major version bump might be needed. If that is the case, we should combine multiple breaking changes at once. (Releasing, not this PR)

sea-bass:
> * The ROS2 example, doesn't contain setting parameters. Is this still possible? In case it is, could we add it to the example?

I think it's still possible though the `rosapi_node` executable in `rosbridge`. Not something I've tried before, but let me do that right now.

> * The changes in the rosbridge_server aren't released yet, correct? Therefore we can merge this one, but we shouldn't release it yet. Correct?
> * This is not really a breaking change, but it only works with the ROS2 rosbridge_server correct? So therefore a major version bump might be needed. If that is the case, we should combine multiple breaking changes at once. (Releasing, not this PR)

Correct -- nothing is released yet, so we should wait until it's all merged in together for ROS 2.

---

BTW for those using TypeScript, like ourselves, I also had to add these changes in support of this PR: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67407

How does this play with having separate releases/branches between `master` and `ros2`?

sea-bass:
Update: Params did work, and I've added an example into the `ros2_simple.html` demo.

MatthijsBurgh:
@sea-bass I think I want to make one final release for v1.X. I will rename the master branch to v1. I will also create a v2 branch, which acts as the release branch for all future v2.x releases.

After which we should rebase this PR on the current develop branch. As the ros2 branch is just behind the develop branch. So we should also merge this PR into the develop branch.

sea-bass:
> @sea-bass I think I want to make one final release for v1.X. I will rename the master branch to v1. I will also create a v2 branch, which acts as the release branch for all future v2.x releases.
> 
> After which we should rebase this PR on the current develop branch. As the ros2 branch is just behind the develop branch. So we should also merge this PR into the develop branch.

@MatthijsBurgh Seems like @EzraBrooks got https://github.com/RobotWebTools/roslibjs/pull/649 to pass -- do you want that to possibly be the first PR of this v2, and then I'll rebase this branch on that + make sure it's TypeScript compliant?

src/core/ActionResult.js
```diff
@@ -0,0 +1,23 @@
...
+
+var assign = require('object-assign');
+
+/**
+ * An ActionResult is returned from sending a ROS 2 action goal.
+ *
+ * @constructor
+ * @template TResult
+ */
+class ActionResult {
```
EzraBrooks: Is there anywhere using an `instanceof` comparison on this, or can we remove these weird stub types in favor of just letting the user's generic type arguments pass all the way through? I don't really like this model that `Message` set up, but didn't want to API-break it in adopting TypeScript.

Same question about ActionGoal. I think these classes are unnecessary.

sea-bass: The place they are used is in snippets like these: https://github.com/RobotWebTools/roslibjs/blob/ros2-actions/src/core/Action.js#L72-L74

Would we be able to just directly use the `TResult` generic here (and in equivalent places) without doing `new ActionResult(message.values)`?

EzraBrooks: Yeah, we can just pass `message.values` through. This constructor is basically just doing a copy of `message.values`.

EzraBrooks:
Can we move 8b83120 and f77ef8a to a different branch so we can squash this and not have this commit make seemingly-unrelated changes?

sea-bass:
> Can we move [8b83120](https://github.com/RobotWebTools/roslibjs/commit/8b831201aeac93b0e856094d25b549f5934983fc) and [f77ef8a](https://github.com/RobotWebTools/roslibjs/commit/f77ef8a9ac54da1ac87dc6f18280a12e0924f819) to a different branch so we can squash this and not have this commit make seemingly-unrelated changes?

Done!

src/core/Action.js
```diff
@@ -0,0 +1,224 @@
...
+      this.ros.on(id, (message) => {
+        if (message.op === 'cancel_action_goal' && typeof this._cancelCallback === 'function') {
+          this._cancelCallback(id);
+        }
+      });
+    }
+
+    // Call the goal execution function provided.
+    // @ts-expect-error -- possibly null
+    this._actionCallback(rosbridgeRequest.args, id);
```
EzraBrooks: Let's try not to add any more typescript workarounds now that we're adding new functionality. Does this work?

```suggestion
    if (this._actionCallback !== null) {
      this._actionCallback(rosbridgeRequest.args, id);
    }
```

EzraBrooks (changes requested):
LGTM other than one question about not avoiding the type checker

src/core/Action.js
sea-bass: Good call, to be consistent with other code in this library I checked that its type was `'function'` rather than null, and also patched up a few more things related to stripping away this comment. Let me know!

EzraBrooks (approved):
Code LGTM and we've been using this successfully in our project!

sea-bass:
@MatthijsBurgh would you like to give this another look before we merge?

All seems to work great for us, except that since we got rid of the build output a few days ago we can't `pnpm install` directly from this git branch and need to cut an actual release soon.

MatthijsBurgh (approved):

MatthijsBurgh:
Before creating a release, I suggest we figure out whether we want to provide compiled files attached to the release, etc. I will create an issue to discuss.

tall1:
I'm trying to run the ros2_action_server.html and I'm runnning into this:
`Uncaught ReferenceError: Buffer is not defined`

Any idea how to solve this issue?
It has something to do with `RosLib.umd.cjs` but I'm not sure how to fix this.

sea-bass:
> I'm trying to run the ros2_action_server.html and I'm runnning into this: `Uncaught ReferenceError: Buffer is not defined`
> 
> Any idea how to solve this issue? It has something to do with `RosLib.umd.cjs` but I'm not sure how to fix this.

I think I ran into this at some point and resolved it with https://github.com/RobotWebTools/roslibjs/pull/692. Are you on a version of roslibjs with that commit?

tall1:
Yes I am.

tall1:
I tried a few more things in adjusting the package but it fails to run.
Is there a working example I can use of this module?

I'm on Humble btw.

sea-bass:
I think with these questions going forward, it's best to open a new issue instead of jumping into a closed PR's discussion thread. And if you do, it might be helpful to provide more details on what you've tried and what error(s) you may be getting. It's a bit challenging to help otherwise.

Setup should be done as in https://github.com/RobotWebTools/roslibjs?tab=readme-ov-file#usage. Basically just running `npm install` in this repo root.

Then, there are several examples in this folder: https://github.com/RobotWebTools/roslibjs/tree/develop/examples -- specifically, the ones prefixed with `ros2_`.

When you open those HTML files in a browser, they display instructions for how to start rosbridge in the background and start up certain ROS executables in the background to communicate with the example.

dineshHarikrishnan:
what is the roslib version we need to support Ros2 actions with roslibjs?

MatthijsBurgh:
@dineshHarikrishnan please open a new issue. PR's are not a place to discuss such issues.

Issue #647, opened 2023-11-16
Increase general library quality


MatthijsBurgh:
The [robostack fork](https://github.com/RoboStack/roslibjs), has applied some improvements like:
- Migrating to typescript
- Migrating to ES6 classes
- Replacing grunt by rollup

This might be good improvements to apply to the original library as well. Preferably we include these changes also in the `ros3djs` and `ros2djs` libraries.

EzraBrooks:
Agreed. I would recommend we start by adding a TypeScript compiler configuration that allows JavaScript files, fixing any issues with statically type checking the current JS, then move forward with TS.

EzraBrooks:
Once #665 (and all the dependent PRs that have come out of it) are in, I plan on making a PR to switch from Grunt + Browserify + Karma to Vite (which includes Rollup) and Vitest, as that's what I have experience with using lately and I think Vite's "library mode" is a very sensible preset for a package like roslibjs.

PR #665, opened 2024-01-08
Modernize import syntax to ESModules

EzraBrooks:
**Public API Changes**

This is _definitely a breaking change_. Code requiring CommonJS modules may no longer work the same way or at all.

**Description**

This removes usage of the old CommonJS module `require` syntax in favor of ECMAScript Module `import` syntax, to ease the adoption of a newer bundler like Rollup. This code works in my Vite-based codebase, so it should work in Rollup.

The bulk of this PR is a change to `package-lock.json` from shuffling dependencies around.

Related to desire for a better/newer bundler in #647.

### DIFF
--- a/Gruntfile.cjs
+++ b/Gruntfile.cjs
@@ -1,6 +1,3 @@
-
-'use strict';
-
 module.exports = function(grunt) {
 
   // Time how long tasks take. Can help when optimizing build times
--- a/Gruntfile.cjs
+++ b/Gruntfile.cjs
@@ -1,3 +1,4 @@
+
 'use strict';
 
 module.exports = function(grunt) {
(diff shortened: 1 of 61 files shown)


EzraBrooks:
Had to upgrade from eventemitter2 to eventemitter3 to get this PR to work at all because of broken TypeScript declarations in eventemitter2. Extracted that change to #668.

EzraBrooks:
Sorry for the delay here, it's actually a good sign, because this branch has been working in such a bulletproof manner on my project that I've been forgetting to come back and finish this..

EzraBrooks:
See #687 for why we're doing this! The existing library layout didn't work with Rollup and Vite, but after this PR it does!

MatthijsBurgh (approved):

PR #670, opened 2024-01-25
Migrate from jshint to eslint

EzraBrooks:
**Public API Changes**

None.

**Description**

JSHint is no longer well supported, most tooling plays well with ESLint. I used `polyjuice` to convert the old `.jshintrc` to an `.eslintrc`, and then manually edited it into the new "flat config" `eslint.config.js` syntax.


sea-bass (approved):
Lint is green, I smash approve button

PR #672, opened 2024-01-25
Unify NodeJS and browser implementation

EzraBrooks:
**Public API Changes**

- Removal of raw TCP socket support and NodeJS-stream-based Topic API
- Removal of "RosLibNode" and "RosLibBrowser" in favor of a singular "RosLib" entrypoint.

**Description**

Removing this Node-specific functionality that seems to be largely untouched in almost 10 years allows having one unified entry point into the codebase. This decreases the maintenance burden of the library and opens the door to swapping these transports out at runtime instead of at build-time; with ESM and dynamic imports, that is very doable - but would require a lot of rearchitecture to accommodate a `Promise`-based async API for any functions that would use a dynamic `import` (since that returns a `Promise`)


EzraBrooks:
This is passing tests locally for me.. wonder why not in CI

EzraBrooks:
This is yet another thing that has become a hurdle in #665, since `import` statements are evaluated at script load time, unlike `require` (which is just a function call).

sea-bass (approved):

PR #683, opened 2024-02-20
extend eslint config to tests/misc. JS files

EzraBrooks:
**Public API Changes**

None.

**Description**

Extends our ESLint config to all script files in the repository and fixes issues arising from that.

Removes the `new-cap` rule from our ESLint config because we violate it in our own API; `Ros.Topic`, `Ros.Service`, etc. are all non-compliant with that rule.


sea-bass (approved):
double quotes begone

PR #691, opened 2024-03-01
Remove Message, ServiceRequest, and ServiceResponse types

sea-bass:
**Public API Changes**
This gets rid of `ROSLIB.Message`, `ROSLIB.ServiceRequest`, and `ROSLIB.ServiceResponse` since those data types were simple wrappers that had little value. Now we can use dictionaries as is (or whatever they're called in JavaScript).

**Description**


src/actionlib/SimpleActionServer.js
```diff
@@ -83,10 +82,8 @@ export default class SimpleActionServer extends EventEmitter {
         // needs to happen AFTER rest is set up
         that.emit('cancel');
       } else {
-        // @ts-expect-error -- we need to design a way to handle arbitrary fields in Message types.
-        that.statusMessage.status_list = [
-          { goal_id: goalMessage.goal_id, status: 1 }
-        ];
+        // @ts-ignore Don't know how to fix this
+        that.statusMessage.status_list = [{ goal_id: goalMessage.goal_id, status: 1 }];
```
sea-bass: @EzraBrooks any pointers to actually fix this one?

EzraBrooks (approved):
As far as I can tell, these were useless and just getting in the way of making typescript types (because you seemingly can't make an interface that extends a non-static type like a generic type argument)

PR #692, opened 2024-03-01
Import fixes in demos

sea-bass:
**Description**
I fixed up the HTML demos since they needed to include a different file with recent changes.


EzraBrooks (approved):
Thanks! I missed doing the find-and-replace here when I modified package.json accordingly.

PR #698, opened 2024-03-18
Add `advertiseAsync` method to `Service`

EzraBrooks:
**Public API Changes**

Adds an `advertiseAsync` function to the `Service` class to allow Promise-returning functions as callbacks to services.

**Description**

☝🏻 
This is useful for cases where, like a feature I'm working on right now in my application, you want to wait to return a Service call until a different asynchronous operation completes. The original `advertise` function is very limiting because it requires all operations undertaken by the Service server to be synchronous due to the design of asynchronicity in JavaScript disincentivizing synchronizing async closures into synchronous ones.

### DIFF
--- a/src/core/Service.js
+++ b/src/core/Service.js
@@ -126,4 +126,37 @@ export default class Service extends EventEmitter {
 
     this.ros.callOnConnection(call);
   }
+
+  /**
+   * An alternate form of Service advertisement that supports a modern Promise-based interface for use with async/await.
+   * @param {(request: TRequest) => Promise<TResponse>} callback An asynchronous callback processing the request and returning a response.
+   */
+  advertiseAsync(callback) {
+    if (this.isAdvertised) {
+      throw new Error('Cannot advertise the same Service twice!');
+    }
+    this.ros.on(this.name, async (rosbridgeRequest) => {
+      /** @type {{op: string, service: string, values?: TResponse, result: boolean, id?: string}} */
+      let rosbridgeResponse = {
+        op: 'service_response',
+        service: this.name,
+        result: false
+      }
+      try {
+        rosbridgeResponse.values = await callback(rosbridgeRequest.args);
+        rosbridgeResponse.result = true;
+      } finally {
+        if (rosbridgeRequest.id) {
+          rosbridgeResponse.id = rosbridgeRequest.id;
+        }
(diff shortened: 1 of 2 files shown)


vite.config.js
```diff
@@ -33,6 +33,7 @@ export default defineConfig({
   test: {
     include: [
       'src/**\/*.{test,spec}.?(c|m)[jt]s?(x)',
+      'test/**\/*.{test,spec}.?(c|m)[jt]s?(x)',
```
EzraBrooks: move to another PR

src/core/Service.js
```diff
@@ -126,4 +126,37 @@ export default class Service extends EventEmitter {
...
   }
+
+  /**
+   * An alternate form of Service advertisement that supports a modern Promise-based interface for use with async/await.
+   * @param {(request: TRequest) => Promise<TResponse>} callback An asynchronous callback processing the request and returning a response.
+   */
+  advertiseAsync(callback) {
+    if (this.isAdvertised) {
+      throw new Error('Cannot advertise the same Service twice!');
+    }
```
EzraBrooks, line 137: I think we should move toward this behavior in all of the cases throughout the library where there's currently guard conditions that just early-return with no information. Users shouldn't be calling `advertise` multiple times on the same `Service` object, that's probably an issue with their code!

I don't want to break that in existing functions yet, but figured I'd do it here where we have a green field.

sea-bass (approved):
very nice!

src/core/Service.js
MatthijsBurgh: We have not released any v2 version yet. So I think we can fix it everywhere.

MatthijsBurgh (commented):
Do we to add unadvertise functionality?

EzraBrooks:
I believe the existing `unadvertise` should work here - note that the existing `unadvertise` doesn't actually remove the event handler from the EventEmitter, so no changes are needed to make this behave the same. That's actually probably a bug - maybe worth fixing for both the synchronous and asynchronous version at the same time.

MatthijsBurgh:
> That's actually probably a bug - maybe worth fixing for both the synchronous and asynchronous version at the same time.

@EzraBrooks sounds like a good idea

EzraBrooks:
I'll merge this and open up a follow-up PR that fixes `unadvertise` for both cases (and add a relevant unit test).

PR #701, opened 2024-03-19
Fix un-advertisement and multiple-advertisement of Service callbacks

EzraBrooks:
**Public API Changes**

Calling `Service.advertise` multiple times now throws an error, as suggested in #698.
Bugfix: prevent resource leak of Service callbacks

**Description**

As noted in #698, Service un-advertisement apparently has never removed the callback from the service server, which is probably not good.. for both resource leak reasons and possible unexpected behavior reasons like a callback being invoked even after unadvertisement.

### DIFF
--- a/src/core/Service.js
+++ b/src/core/Service.js
@@ -24,6 +24,10 @@ export default class Service extends EventEmitter {
     this.serviceType = options.serviceType;
     this.isAdvertised = false;
 
+    /**
+     * Stores a reference to the most recent service callback advertised so it can be removed from the EventEmitter during un-advertisement
+     * @type {((rosbridgeRequest) => any) | null}
+     */
     this._serviceCallback = null;
   }
   /**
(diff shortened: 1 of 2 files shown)


src/core/Service.js
```diff
@@ -85,46 +89,51 @@ export default class Service extends EventEmitter {
...
-    if (this.isAdvertised || typeof callback !== 'function') {
-      return;
+    if (this.isAdvertised) {
+      throw new Error('Cannot advertise the same Service twice!');
     }
 
-    this._serviceCallback = callback;
-    this.ros.on(this.name, this._serviceResponse.bind(this));
+    // Store the new callback for removal during un-advertisement
+    this._serviceCallback = (rosbridgeRequest) => {
```
MatthijsBurgh, line 97: In python I am really against adding class variables outside the constructor. I don't know how you feel about it in JS/TS.

MatthijsBurgh (approved):
Only one comment, for the rest I approve.

src/core/Service.js
EzraBrooks: This class variable is declared in the constructor, just set to the actual callback here. I don't think TypeScript even lets you add class members that weren't declared in the constructor or in the class body :)

This reminds me that I should go around and clean up all the places where class fields are being initialized with constants (not constructor parameters) and move them out of the constructor into the body of the class - I'll do that here first.

PR #731, opened 2024-05-29
Add `ROS2TFClient`

pac48:
This PR adds a `ROS2TFClient` class to interact with the ROS 2 tf2_web_republisher.

### DIFF
--- a/src/tf/TF2Client.js
+++ b/src/tf/TF2Client.js
@@ -4,9 +4,6 @@
  */
 
 import Action from '../core/Action.js';
-// import Goal from '../actionlib/Goal.js';
-// import Topic from '../core/Topic.js';
-
 import Transform from '../math/Transform.js';
 
 import Ros from '../core/Ros.js';
@@ -39,15 +36,12 @@ export default class TF2Client extends EventEmitter {
         this.updateDelay = options.updateDelay || 50;
         var seconds = options.topicTimeout || 2.0;
         var secs = Math.floor(seconds);
-        var nsecs = Math.floor((seconds - secs) * 1000000000);
+        var nsecs = Math.floor((seconds - secs) * 1E9);
         this.topicTimeout = {
             secs: secs,
             nsecs: nsecs
         };
         this.serverName = options.serverName || '/tf2_web_republisher';
-        // this.repubServiceName = options.repubServiceName || '/republish_tfs';
-
-        // this.currentGoal = {};
         this.goal_id = '';
         this.frameInfos = {};
         this.republisherUpdateRequested = false;
@@ -184,8 +178,5 @@ export default class TF2Client extends EventEmitter {
      */
     dispose() {
         this._isDisposed = true;
-        // if (this.currentTopic) {
-        //     this.currentTopic.unsubscribe(this._subscribeCB);
-        // }
     }
 }
--- a/src/tf/TF2Client.js
+++ b/src/tf/TF2Client.js
@@ -1,7 +1,3 @@
-/**
- * @fileOverview
- * @author David Gossow - dgossow@willowgarage.com
- */
 
 import Action from '../core/Action.js';
 import Transform from '../math/Transform.js';
--- a/src/tf/TF2Client.js
+++ b/src/tf/TF2Client.js
@@ -1,0 +1,191 @@
+/**
+ * @fileOverview
+ * @author David Gossow - dgossow@willowgarage.com
+ */
+
+import Action from '../core/Action.js';
+// import Goal from '../actionlib/Goal.js';
+// import Topic from '../core/Topic.js';
+
+import Transform from '../math/Transform.js';
+
+import Ros from '../core/Ros.js';
+import {EventEmitter} from 'eventemitter3';
+
+/**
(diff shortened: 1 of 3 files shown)


src/tf/TF2Client.js
```diff
@@ -0,0 +1,182 @@
...
+import Action from '../core/Action.js';
+import Transform from '../math/Transform.js';
+
+import Ros from '../core/Ros.js';
+import {EventEmitter} from 'eventemitter3';
+
+/**
+ * A TF Client that listens to TFs from tf2_web_republisher.
+ */
+export default class TF2Client extends EventEmitter {
```
EzraBrooks: I think this needs to be called something like `ROS2TFClient`, since "TF2" works on both ROS1 and ROS2.

src/tf/TF2Client.js
```diff
@@ -0,0 +1,182 @@
+/**
+ * @fileOverview
+ * @author David Gossow - dgossow@willowgarage.com
+ */
```
EzraBrooks: lol
```suggestion
```

src/tf/TF2Client.js
```diff
@@ -0,0 +1,182 @@
...
+            trans_thres: this.transThres,
+            rate: this.rate
+        };
+
+        if (this.goal_id !== '') {
+            this.actionClient.cancelGoal(this.goal_id);
+        }
+        this.currentGoal = goalMessage;
+
+        // this.currentGoal.on('feedback', this.processTFArray.bind(this));
```
EzraBrooks: remove commented code

src/tf/TF2Client.js
```diff
@@ -0,0 +1,182 @@
...
+        const id = this.actionClient.sendGoal(goalMessage,
+            (result) => {
+                console.log('Result for action goal on :');
```
EzraBrooks: let's not pollute the logs with debug info

EzraBrooks (changes requested):
Thanks for working on this! Can you explain the API difference between your ROS 2 version of tf2_web_republisher and the ROS 1 version? If we can figure out a way to unify them (instead of renaming this `ROS2TFClient` or similar) that would be ideal!

src/tf/TF2Client.js
```diff
@@ -0,0 +1,182 @@
...
+        super();
+        this.ros = options.ros;
+        this.fixedFrame = options.fixedFrame || 'base_link';
+        this.angularThres = options.angularThres || 2.0;
+        this.transThres = options.transThres || 0.01;
+        this.rate = options.rate || 10.0;
+        this.updateDelay = options.updateDelay || 50;
+        var seconds = options.topicTimeout || 2.0;
+        var secs = Math.floor(seconds);
+        var nsecs = Math.floor((seconds - secs) * 1E9);
```
Hoffalypse: nitpicky but change these var to const

EzraBrooks (approved):
LGTM as a first cut. Naming it ROS2TFClient makes extra certain that we won't accidentally step on ROS 1 users' toes. Later on it may make sense to combine the two classes and use a flag, or something like that.

EzraBrooks:
For posterity, the "ROS 2 version of tf2_web_republisher" we are currently working against is @pac48's fork of it at https://github.com/pac48/tf2_web_republisher. We're still deciding where this implementation should live long-term; it may become a ros2 branch of the RobotWebTools project that is largely currently defunct.

pfarinha91:
Thanks for your work on this TF2 client for ROS2! 

I'm trying to make this approach work on a react application, but I'm struggling. I stumbled upon this thread and [this one](https://github.com/RobotWebTools/ros3djs/issues/658), changed the Roslibjs version on package.json to `github:pac48/roslibjs#pr-add-TF2Client`. Installed it and the rosbridge complained about not having the `send_action goal` operation. Basically actions are still not implemented on the main release of rosbridge for Ros Humble, which is behind the repository. So I cloned and compiled the most recent version of rosbridge, launched the websocket plus the tf2_web_republisher fork from @pac48 and I finally can subscribe to TFs and they start arriving through the callback! But...

**I can't unsubscribe or dispose the ROS2TFClient** when the parent React component unmounts (or in any moment whatsoever). Worst, the moment I subscribe to a new frame using the client, I can no longer unsubscribe from any topic (even in normal subscribers) and eventually `WebSocketClosedError: Tried to write to a closed websocket` errors start raining on the rosbridge websocket if I try to reconnect.

EzraBrooks:
Hi, that's interesting - I'm not experiencing the same symptom. Could you start a Discussion on this topic? I can provide some insight there as to integrating this with React (as my team is also using it with React), but would rather have that conversation in a Discussion than a PR thread so we can "mark as answer" when we find the solution :)

EzraBrooks:
I just realized Discussions weren't enabled here - I've enabled them!

pfarinha91:
Thanks! Are you talking about discourse.ros.org? When I try to create a Discussion it redirects there.

Edit: nevermind, thay have appeared now!

pfarinha91:
I've created the Discussion topic on https://github.com/RobotWebTools/roslibjs/discussions/742

PR #760, opened 2024-07-17
Add a `getPublishers` method to the Topic class

hello-amal:
**Public API Changes**

Add a `getPublishers` method to the `Ros` class, which returns all publishers for a topic.

**Description**

`rosapi` in `rosbridge` [exposes a `/rosapi/publishers` service to get all the publishers for a topic](https://github.com/RobotWebTools/rosbridge_suite/blob/d22f102b59e7d9fdeea0ec5e74aa8b98358585d7/rosapi/scripts/rosapi_node#L107). This change exposes that functionality to the ROSlib client.

Note: I tested this in a separate fork with ROS2, which is what I use. I haven't tested it in ROS1 -- hopefully the CI is sufficient for that.


EzraBrooks (commented):
What would you think about moving this into the `Topic` class instead?

EzraBrooks (approved):

PR #761, opened 2024-07-19
(README) Explain where built files can be found

dj1yfk:
**Public API Changes**
None

**Description**
To users not familiar with npm, it may not be obvious what to do after installing the package with npm. Therefore a small explanation is added, where the built files can be found.


MatthijsBurgh:
To users that are not familiar with the NPM/yarn eco system, they should refer to their relevant documentation. This is not the place to document it.

PR #762, opened 2024-07-22
Fix undefined reference to "WebSocket" for NodeJS instances

### LINKED ISSUE #756: Node.js: ReferenceError: WebSocket is not defined
**Description**

The shim logic for deciding which WebSocket library to load runs _after_ some code that tries to use the browser-specific `WebSocket` global:
https://github.com/RobotWebTools/roslibjs/blob/39158ec5fdb8ad540461ad92f2447ecdaef4757c/src/core/Ros.js#L64

This, in Node.js, then results in the error in the title -- `ReferenceError: WebSocket is not defined`.

- Library Version: https://github.com/RobotWebTools/roslibjs/commit/7657e2bfebc93faa860e121ad10f13d70c517bd3
- ROS Version: Iron
- Platform / OS: Linux

**Steps To Reproduce**
* Create a connection
* Try to connect without ROS ready/running yet, so the first attempt fails and `this.socket` is set but it's not connected
* Try to reconnect after ROS is ready
* :boom:

**Expected Behavior**
It shouldn't crash

**Actual Behavior**
It crashes

**Possible Fix**
Since the value of this constant appears to be set by the WhatWG spec (https://websockets.spec.whatwg.org/#dom-websocket-closed), I think it can just be replaced with `... === 3` and an appropriate comment.

Alternately, the nesting order of the different `if` conditions could be changed so it checks `... === ws.WebSocket.CLOSED` on the Node.js branch.

ghost:
**Public API Changes**
None

**Description**
Fixed an undefined reference to `WebSocket` in src/core/Ros.js.  This is a fairly direct approach. We could also rearrange the logic to reference `ws.WebSocket.CLOSED` on the NodeJS branch. see the issue.

Resolves #756

### DIFF
--- a/src/core/Ros.js
+++ b/src/core/Ros.js
@@ -61,7 +61,8 @@ export default class Ros extends EventEmitter {
         socketAdapter(this)
       );
     } else if (this.transportLibrary === 'websocket') {
-      if (!this.socket || this.socket.readyState === WebSocket.CLOSED) {
+      // Use value of `WebSocket.CLOSED`, 3 to avoid invalid reference for NodeJS instances https://websockets.spec.whatwg.org/#dom-websocket-closed
+      if (!this.socket || this.socket.readyState === 3) {
         // Detect if in browser vs in NodeJS
         if (typeof window !== 'undefined') {
           const sock = new WebSocket(url);
--- a/src/core/Ros.js
+++ b/src/core/Ros.js
@@ -61,21 +61,22 @@ export default class Ros extends EventEmitter {
         socketAdapter(this)
       );
     } else if (this.transportLibrary === 'websocket') {
-      // Use value of `WebSocket.CLOSED`, 3 to avoid invalid reference for NodeJS instances https://websockets.spec.whatwg.org/#dom-websocket-closed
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
@EzraBrooks do we want to solve it this way, or do we need to import the [websocket class](https://github.com/websockets/ws/blob/HEAD/doc/ws.md#class-websocket) here?

EzraBrooks (commented):
Thanks for the PR, @mhosmar-cpr! I missed this bug while writing this shim because we don't have unit tests that cover this code path - do you have time to add one?

I also think that we should probably move this condition one layer down as you and @MatthijsBurgh described - even though it's duplicative, it at least is super clear where the magic number comes from if we do that.

ghost:
I have implemented the change as described. However I am not sure of the potential cost of repeatedly importing `ws`. If it does not pose any risk I think this change is good as is.

Unfortunately I don't have the knowledge or time to implement a unit test for this code path. Please understand.

EzraBrooks (approved):

ghost:
looping back. FYI I can't merge even with approval.

EzraBrooks:
Running CI and will merge once it passes

PR #796, opened 2024-09-29
Add status field to action result messages

sea-bass:
**Public API Changes**
Adds a new `Action.setCanceled()` methods to properly cancel actions server-side using `roslibjs`.

**Description**
This goes along with https://github.com/RobotWebTools/rosbridge_suite/pull/953, in which a `status` field was added to the `action_result` message.

It also adds a `setCanceled()` helper function to the `Action` class to wrap the setting of the right status in cancellation settings.


EzraBrooks (commented):
I'm like half asleep and on a plane right now so maybe I'm missing context, but I'm not sure I grok what's going on here. Is this an action-server-side feature or action-client-side?

sea-bass:
> I'm like half asleep and on a plane right now so maybe I'm missing context, but I'm not sure I grok what's going on here. Is this an action-server-side feature or action-client-side?

This is a server-side feature.

For context, take a look at https://github.com/RobotWebTools/rosbridge_suite/issues/920 and the PR that tries to solve it (https://github.com/RobotWebTools/rosbridge_suite/pull/953).

Basically, there are 2 new things:

1. Now the `action_result` message carries a status with it (hence the additions to the `setSucceeded()`/`setFailed()` functions here)
2. Also, when you cancel a goal client-side, instead of the client automatically aborting the goal, the server is responsible for telling you that it was canceled (hence the addition of the `setCanceled()` function here).

sea-bass:
I fully dropped the ball on re-requesting review. Now that rosbridge 2.x is released, this is needed for action support to be compatible.

src/core/Action.js
```diff
@@ -216,6 +216,25 @@ export default class Action extends EventEmitter {
       id: id,
       action: this.name,
       values: result,
+      status: 4,  // Corresponds to GoalStatus.STATUS_SUCCEEDED
```
EzraBrooks: Can we extract GoalStatus to an enum, since we've got typescript?

EzraBrooks (changes requested):

src/core/Action.js
sea-bass: Apparently since these are `.js` files, things are not happy if I try it directly:

```
src/core/Action.js:14:6 - error TS8006: 'enum' declarations can only be used in TypeScript files.
```

Is there some easy way around this that fits within this PR?

sea-bass: I found this... please let me know if there's anything better.

https://stackoverflow.com/a/44447975

EzraBrooks: Did you try adding a typescript file for the enum and importing it into the JS file? I don't know in practice how that works, but it should transpile into an object?

sea-bass: Yeaaaa, that did work! Just had to add one more compiler option. Thank you.

tsconfig.json
```diff
@@ -5,6 +5,8 @@
     "rootDir": "./src" /* Specify the root folder within your source files. */,
     "allowJs": true /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */,
     "checkJs": true /* Enable error reporting in type-checked JavaScript files. */,
+    "allowImportingTsExtensions": true /* Enable importing TypeScript files. */,
+    "emitDeclarationOnly": true /* Needed to allow importing TypeScript files. */,
```
EzraBrooks: Are you sure these two are needed? I thought `allowImportingTsExtensions` just lets you put `.ts` in the filename instead of excluding the extension from the import.

sea-bass: Without the first one, my `npm run build` failed.

The second one was simply suggested by my linter in VSCode but did not seem necessary.

sea-bass: Oh, you mean that I could omit the extension from the import and it may work? I will try when I get a chance

sea-bass: I attempted this and have an alternate error:

```
Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'.
Did you mean '../core/GoalStatus.js'?ts(2835)
```

Indeed, the `tsconfig.json` has `"moduleResolution": "nodenext"` specified.

I could get around this by switching the settings from

```json
    "module": "nodenext",
    "moduleResolution": "nodenext",
```

to

```json
    "module": "esnext",
    "moduleResolution": "bundler",
``` 

... but to be honest with you, I have no idea what I'm touching at this point. Maybe there are undesired consequences.

EzraBrooks (approved):
I think now that we're building roslibjs with Vite, `nodenext` and `bundler` actually are the right settings for the tsconfig. I seem to recall the previous settings being a side effect of the odd halfway step we were in a year or so ago before we adopted a fully modern toolchain.

sea-bass:
> I think now that we're building roslibjs with Vite, `nodenext` and `bundler` actually are the right settings for the tsconfig. I seem to recall the previous settings being a side effect of the odd halfway step we were in a year or so ago before we adopted a fully modern toolchain.

Confirming since right now this changes to `esnext` and `bundler` -- as `nodenext` did not allow importing without extensions.

EzraBrooks:
sorry yea I meant `esnext`. I am half-braindead because yesterday I drove from Colorado to the east coast

PR #823, opened 2024-12-05
fix(params): align parameter retrieval with rosbridge changes

Mehsias:
**Public API Changes**
None

**Description**
- fixed name and type of rosapi service

Related to RobotWebTools/rosbridge_suite#972

### DIFF
--- a/src/core/Param.js
+++ b/src/core/Param.js
@@ -36,8 +36,8 @@ export default class Param {
   get(callback, failedCallback) {
     var paramClient = new Service({
       ros: this.ros,
-      name: '/rosapi/get_param',
-      serviceType: 'rosapi/GetParam'
+      name: '/rosapi/get_param_names',
+      serviceType: 'rosapi_msgs/srv/GetParamNames'
     });
 
     var request = {name: this.name};


MatthijsBurgh:
This is just wrong. The service type should be `rosapi_msgs/srv/GetParam` and the name should still be the same.

Mehsias:
@MatthijsBurgh can you elaborate? 

Would it be an option to create a new function to retrieve all parameter names?

Mehsias:
ros2 service info /rosapi/get_param yields -> Type: rosapi_msgs/srv/GetParam
 
 in the original code it is Type: rosapi/GetParam, so why is that not an issue?
 
 I get that my initial PR was false, as I should have just created a new function. But now I am curious why this is not an issue?

bjsowa:
> ros2 service info /rosapi/get_param yields -> Type: rosapi_msgs/srv/GetParam
> 
> in the original code it is Type: rosapi/GetParam, so why is that not an issue?
> 
> I get that my initial PR was false, as I should have just created a new function. But now I am curious why this is not an issue?

If you look through the [rosbridge library code](https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/rosbridge_library/src/rosbridge_library/internal/ros_loader.py#L86), you can see that it adds the `msg` or `srv` subnames to type names if they don't already contain it. This makes the roslib library compatible with ROS1 and ROS2.

PR #835, opened 2025-01-20
Add optional timeout argument to callService

bjsowa:
**Public API Changes**
Added optional `timeout` argument to callService.

**Description**
After RobotWebTools/rosbridge_suite#984 was merged, it is now possible to specify a timeout when calling a service. This PR exposes this functionality to roslibjs public API.

### DIFF
--- a/src/core/Service.js
+++ b/src/core/Service.js
@@ -73,12 +73,9 @@ export default class Service extends EventEmitter {
       id: serviceCallId,
       service: this.name,
       type: this.serviceType,
-      args: request
+      args: request,
+      timeout: timeout
     };
-
-    if (timeout !== undefined) {
-      call.timeout = timeout;
-    }
 
     this.ros.callOnConnection(call);
   }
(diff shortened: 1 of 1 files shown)


src/core/Service.js
```diff
@@ -73,6 +75,11 @@ export default class Service extends EventEmitter {
       type: this.serviceType,
       args: request
     };
+
+    if (timeout !== undefined) {
+      call.timeout = timeout;
+    }
+
```
EzraBrooks: Is there a reason that this can't be put in the definition of the object above? Does passing `timeout: undefined` into rosbridge break it? That seems like an API issue, since it can be tricky to easily discern in a JavaScript application whether something is _deliberately undefined_ or _missing_.

I'd rather this read
```suggestion
      args: request,
      timeout: timeout
    };
```

and if this causes an issue in rosbridge we should fix it there.

EzraBrooks (commented):
This is an awesome addition, I have a client-side hack around the lack of this in my application. I have an implementation question though

src/core/Service.js
bjsowa: It seems that properties with undefined value are just being omitted when converted to JSON.

EzraBrooks (approved):

PR #836, opened 2025-01-20
Use relative names for rosapi services

bjsowa:
**Public API Changes**
None

**Description**
When rosbridge server is running in a namespace, we can expect rosapi services to be in the same namespace (after RobotWebTools/rosbridge_suite#992 was merged). To contact services that are in the same namespace as the rosbridge server, we can just use relative topic names (without leading slash) instead of global ones.

This will help running rosbridge in multi-robot scenarios, where each robot has its own rosbridge server and rosapi nodes running under the robot's namespace.

### DIFF
--- a/src/core/Param.js
+++ b/src/core/Param.js
@@ -36,7 +36,7 @@ export default class Param {
   get(callback, failedCallback) {
     var paramClient = new Service({
       ros: this.ros,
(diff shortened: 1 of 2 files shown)


EzraBrooks (approved):
LGTM

PR #842, opened 2025-02-05
Add namespace support for the new version of Rosbridge_suite

FlorianLebecque:
**Public API Changes**
Add an optional namespace parameter to the Ros class

**Description**
The roslibjs is broken when setting a namespace to the rosbriduge_suite since [this PR on rosbridge_suite](https://github.com/RobotWebTools/rosbridge_suite/pull/992).
This change allows dev to set a namespace on the Ros object. This parameter is optional and by default an empty string. When populated, act as a prefix for the different ROSAPI services. 

```js
const ros = new ROSLIB.Ros({
    url: props.url
});

ros.namespace = "NAME_SPACE_IN_ROS/";
```

this allows the Roslibjs to call "NAME_SPACE_IN_ROS/rosapi/Topics" instead of "rosapi/Topics"


MatthijsBurgh:
The ROS class doesn't have a namespace attribute. So this should be set in the constructor. Also tests should be added so it works with absolute, relative and private namespaces. And I think it requires more parsing than just prefixing the topics/services with the namespace. As you want to be robust against a missing slash at the end of the namespace.

EzraBrooks:
I'd also appreciate the use of template strings instead of string concatenation here for readability :)

FlorianLebecque:
I've done change to add the namespace to the constructor.
I've added better logic for 'parsing' of the namespace. this should make it more robust to missing '/' and leading '/'

Unfortunately, I have no idea how to test it, I'am not familiar with javascript testing. Do you have any ressource I could use to add it?

Since testing it requires having multiple rosbridge_suite with different namespaces running on different port for the test.

test/namespaces.test.js
```diff
@@ -0,0 +1,128 @@
...
+
+        const topics = await new Promise((resolve, reject) => {
+            ros.getTopics(resolve, reject);
+        });
+
+        expect(topics).be.an('array');
+        expect(topics).not.hasLength(0);
+    })
+
+})
```
MatthijsBurgh: Missing new line at EOF

src/core/Ros.js
```diff
@@ -29,19 +29,27 @@ export default class Ros extends EventEmitter {
...
     options = options || {};
     this.transportLibrary = options.transportLibrary || 'websocket';
     this.transportOptions = options.transportOptions || {};
     this.groovyCompatibility = options.groovyCompatibility ?? true;
+    this.namespace = options.namespace || '';
+
+    // Normalize namespace format: no leading slash, with trailing slash
+    if (this.namespace) {
+      this.namespace = this.namespace.replace(/^\//, '').replace(/\/?$/, '/');
+    }
```
MatthijsBurgh: Maybe people want to set an absolute namespace...

MatthijsBurgh (changes requested):

src/core/Param.js
```diff
@@ -36,11 +36,11 @@ export default class Param {
   get(callback, failedCallback) {
     var paramClient = new Service({
       ros: this.ros,
-      name: 'rosapi/get_param',
+      name: this.ros.namespace + 'rosapi/get_param',
```
MatthijsBurgh: Use template string here as well

MatthijsBurgh (changes requested):

MatthijsBurgh:
@FlorianLebecque check CI errors

FlorianLebecque:
@MatthijsBurgh How can I setup the test to connect to different rosbridge_suite websockets ? To test different namespace configuration ?

bjsowa (commented):
The roslibjs, after #836, uses relative service names (without leading `/`), thus if rosbridge_server is run under a namespace `<NAMESPACE>`, the rosapi service names will resolve to this namespace. For example `rosapi/get_param` will resolve to `/<NAMESPACE>/rosapi/get_param`. We typically expect rosapi node to be run in the same namespace as rosbridge_server.

MatthijsBurgh:
@bjsowa How does roslibjs namespace resolving? Because it is not a classic node, so it doesn't have node name, so no namespace

FlorianLebecque:
Then this pull request is no longer necessary and can be discarded :)

bjsowa:
> @bjsowa How does roslibjs namespace resolving? Because it is not a classic node, so it doesn't have node name, so no namespace

ROSbridge server acts as the middle man when talking to the ROS world. If the server node is run under a namespace (let's say `ns1` so the fully qualified name of the node resolves to `/ns1/rosbridge_server`), then when roslibjs tries to advertise a service using a relative name (no leading slash), the server will apply the namespace when resolving the name. For example `rosapi/get_param` gets resolved to `/ns1/rosapi/get_param`

EzraBrooks:
closing per @FlorianLebecque's comment above

PR #874, opened 2025-04-18
Support for Deno & Bun

jeff-hykin:
**Public API Changes**
None

**Description**
The only functional change is a conditional check for the WebSocket global. Roslibjs effectively already supports Deno and Bun, because both implement the browser specification of WebSockets. The only thing causing it to fail is that Roslibjs incorrectly assumes theres no WebSocket (it assumes NodeJS) when there's no window object.

The only other change (appending `.ts`) is because Deno natively supports typescript, so long as the full file path is specified.

E.g. the following now works, and it previously did not:

```sh
cd roslibjs
deno
>import * as ros from "./src/RosLib.js"
```


MatthijsBurgh:
The CI is failing, so review is on hold till CI is fixed

jeff-hykin:
@MatthijsBurgh thanks for specifying. It should pass the linter now. 

While an exception to the linter's globals check was maybe justified, it made me realize a more generalized approach for this PR. The revised version is smaller and adds support for Bun, Deno, and possibly other runtimes. 

I should talk to you directly about the other check that now passes (.ts extension). If I could avoid the change I would. Its in this PR because it's a functional difference on runtimes that follow the JS (ECMA Script) spec. Deno implements the browser (ECMA Script) specification with support for typescript. If the import statement is not an exact file path, it fails.

deno.lock
EzraBrooks, line 1: I don't think you need to commit the Deno lockfile, since roslibjs is not shipping via Deno, right?

src/core/Action.js
```diff
@@ -5,7 +5,7 @@
 
 import { EventEmitter } from 'eventemitter3';
 import Ros from '../core/Ros.js';
-import { GoalStatus } from '../core/GoalStatus';
+import { GoalStatus } from '../core/GoalStatus.ts';
```
EzraBrooks, line 8: This makes sense to me but could you double-check the resulting bundled code from Vite to make sure that this resolves okay for runtimes that _don't_ support TypeScript natively like Deno and ts-node do? I assume Vite will handle this, but I'm not 100% certain since I use extensionless paths on my projects.

src/core/Ros.js
```diff
@@ -61,15 +61,15 @@ export default class Ros extends EventEmitter {
         socketAdapter(this)
       );
     } else if (this.transportLibrary === 'websocket') {
-      // Detect if in browser vs in NodeJS
-      if (typeof window !== 'undefined') {
+      // browsers, Deno, and Bun support WebSockets natively
+      if (typeof WebSocket == "function") {
```
EzraBrooks: Please fix the failing checks on this line to adhere to the project's style guidelines

EzraBrooks (commented):
Thanks for the contribution! Some minor comments.

jeff-hykin:
I came to bump this, only to realize it was waiting on me! Sorry for the delay.

Linter passes locally, its now up to date with the latest master, and I confirmed the rename (.js to .ts) is still bundled correctly by vite (screenshot)

Screen Shot 2025-06-04 at 2 32 36 PM

jeff-hykin:
@EzraBrooks just a friendly ping. This 5 line change would really help some of my projects for students.

MatthijsBurgh:
So now it is only one line left?

EzraBrooks (changes requested):
You still need to remove the `deno.lock`. We use npm, putting your Deno lockfile in this repository is going to result in constant desynchronization for Deno users because we won't be updating it upstream.

jeff-hykin:
> You still need to remove the `deno.lock`

I can't believe I missed that again, sorry my brain is so used to skimming over lock files in PRs. It is removed now. I agree it shouldn't be part of the PR (it was already outdated in my branch). 

> So now it is only one line left?

There is only one functional line change for npm, but the other line changes are necessary for the import to succeed in deno (if that is what you're asking). Without the .ts extension the import fails. The line change in the typescript config is to allow the .ts extension.

EzraBrooks (approved):

EzraBrooks (approved):

jeff-hykin:
Thank you! 😄

PR #875, opened 2025-04-18
Add `globalThis`-preserving Import Path

jeff-hykin:
**Public API Changes**
None

**Description**
This has no functional change for all existing users (ex: CDN). RosLib.js exports all the same API's as before and still sets `globalThis.ROSLIB =`.

However, for users like myself who want to avoid polluting `globalThis`, this PR provides`src/index.js`.

NOTE: to avoid code duplication
- `src/index.js` consolidates the imports
- `src/RosLib.js` re-exports everything from index.js, and then additionally sets ROSLIB

### DIFF
--- a/src/RosLib.js
+++ b/src/RosLib.js
@@ -1,28 +1,11 @@
 /**
  * @fileOverview
- * @author Russell Toris - rctoris@wpi.edu
+ * @author Jeff Hykin - jeff.hykin@gmail.com
  */
 
 /** @description Library version */
-export const REVISION = '1.4.1';
-export * from './core/index.js';
-export * from './actionlib/index.js';
-export * from './math/index.js';
-export * from './tf/index.js';
-export * from './urdf/index.js';
+export * from './index.js';
+import ROSLIB from './index.js';
 
-import * as Core from './core/index.js';
-import * as ActionLib from './actionlib/index.js';
-import * as Math from './math/index.js';
-import * as Tf from './tf/index.js';
-import * as Urdf from './urdf/index.js';
-
-// Add to global namespace for in-browser support (i.e. CDN)
-globalThis.ROSLIB = {
-  REVISION,
-  ...Core,
-  ...ActionLib,
-  ...Math,
-  ...Tf,
-  ...Urdf
-};
+// same as index.js, except add to global namespace for in-browser support (i.e. CDN)
+globalThis.ROSLIB = ROSLIB;
--- a/src/RosLib.js
+++ b/src/RosLib.js
@@ -1,6 +1,6 @@
 /**
  * @fileOverview
- * @author Jeff Hykin - jeff.hykin@gmail.com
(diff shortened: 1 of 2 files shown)


EzraBrooks (commented):
Thanks for the cleanup here!

Doesn't this line in `package.json` need to change so that resolving the ESModule resolves this new index.js that doesn't pollute the global namespace?
https://github.com/RobotWebTools/roslibjs/blob/d1c9f8893c4d9bdf303380ba8309deb36ce9e53f/package.json#L11

jeff-hykin:
> Doesn't this line in `package.json` need to change so that resolving the ESModule resolves this new index.js that doesn't pollute the global namespace?

TLDR: I'm proposing something much smaller. I'm trying to make this a no-brainer tiny PR that for-sure breaks nothing.

I'd be happy to open a separate PR for the change you're talking about, which is a much bigger change.

Why does this small change work for me? 
- I import code from URLs rather through npm, so I don't care much about the npm endpoints. I'm directly grabbing the source code.

Why would I make a separate PR for the package.json change?
- I would LOVE to make that change. But to be clear, while rare/unlikely, its possible that change could be breaking for some people who use npm. I'd argue its better to cause that breakage sooner rather than later. But it's a bigger decision. I wanted this PR to be easy to merge before I make bigger PRs.

If there needs to be some debate/discussion/delay around # 2, then I'd ask that this PR be considered/merged as-is (no-breakage) and then I can open up a second (possibly-breaks-things) PR for changing the npm endpoint.

jeff-hykin:
Just wanted to check back in on this since it should be a really small non-breaking change @EzraBrooks

src/RosLib.js
```diff
@@ -1,28 +1,11 @@
 /**
  * @fileOverview
- * @author Russell Toris - rctoris@wpi.edu
+ * @author Jeff Hykin - jeff.hykin@gmail.com
```
sea-bass: Please keep the original author here. You can still add yourself if you like (though personally I would only do that for major changes).

jeff-hykin: Dont worry, I'm perfectly fine with no mention.

However, it is a big deal to me how bad this looks on me (e.g. the diff looks like I edited-out the original author, which is not the story on my end). I renamed RosLib.js to index.js (hence author is maintained there) and then I made a new file. I wasnt sure on your policy for new files so tried to copy the header style and put my info. 

Naturally git doesn't know I renamed RosLib.js to index.js so it looks like I edited the original. I probably should've checked the diff and swapped the headers. 

Sorry for that confusion. Either way, I agree the original author should be at the top of both

sea-bass: Ah yeah, I sort of noticed that when the "new file" in the diff had the former author's name. All good!

jeff-hykin: Alright fixed!

sea-bass (approved):
Thank you!

jeff-hykin:
Thank you for merging!

PR #870, opened 2025-04-08
fixed xmlDoc type error

ygao9795:
fixed this error:

```
src/urdf/UrdfModel.js:35:7 - error TS2322: Type 'Element | null' is not assignable to type 'Element | null | undefined'.
  Type 'Element' is missing the following properties from type 'Element': classList, className, clientHeight, clientLeft, and 102 more.

35       xmlDoc = parser.parseFromString(string, MIME_TYPE.XML_TEXT).documentElement;
         ~~~~~~


Found 1 error in src/urdf/UrdfModel.js:35
```
Add type assertion for xmldom Element compatibility
```
// Check if we are using a string or an XML element
    if (string) {
      // Parse the string
      var parser = new DOMParser();
      var parsedDoc = parser.parseFromString(string, MIME_TYPE.XML_TEXT);
      if (parsedDoc && parsedDoc.documentElement) {
        xmlDoc = parsedDoc.documentElement;
      } else {
        throw new Error('Failed to parse URDF string');
      }
    }
    if (!xmlDoc) {
      throw new Error('No URDF document parsed!');
    }
```

Issue #838, opened 2025-01-22
Importing roslib in Vite application crashes in production as 'this' is undefined


olehakon:
**Description**
I'm using Vite to build a React SPA, hosted through a Node Express server. 

Importing roslib in dev mode works, but after building the application I have an error in the browser saying `Cannot read properties of undefined (reading 'ROSLIB')`. 

The error seems to come from the RosLib.js file, specifically this code:
```
var ROSLIB = this.ROSLIB || {  // <-- Error here
   REVISION : '1.4.1'
};
```

It seems that 'this' is undefined.

Am I missing something here or is this module not compatible with Vite?

- Library Version: 1.4.1
- ROS Version: ?
- Platform / OS: OSX

**Steps To Reproduce**
1. Create a new Vite project
2. Install roslib
3. import roslib, e.g. `import * as ROS from "roslib"`
4. Build the application for production and run it

**Expected Behavior**
Application running as expected

**Actual Behavior**
Application crashes with the follwing error:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'ROSLIB')
```

MatthijsBurgh:
You probably need to build the library from source. As the current HEAD of the library is refactored a lot compared to the latest release. Vite is now used to build the library.

olehakon:
> You probably need to build the library from source. As the current HEAD of the library is refactored a lot compared to the latest release. Vite is now used to build the library.

Ahh, ok. Are there any step-by-step guides how to do this the correct way?
And, are there any plans for a next release with the current refactored codebase?

MatthijsBurgh:
That depends on the package manager you are using. It probably has a guide to add git dependencies.

@sea-bass @EzraBrooks can probably tell you more about the release plans.

olehakon:
> That depends on the package manager you are using. It probably has a guide to add git dependencies.
> 
> [@sea-bass](https://github.com/sea-bass) [@EzraBrooks](https://github.com/EzraBrooks) can probably tell you more about the release plans.

Ok. I'm using yarn and it seems that I can install it directly from github. But when I try that I have an error upon build:
```
src/urdf/UrdfModel.js:35:7 - error TS2322: Type 'Element | null' is not assignable to type 'Element | null | undefined'.
  Type 'Element' is missing the following properties from type 'Element': classList, className, clientHeight, clientLeft, and 102 more.

35       xmlDoc = parser.parseFromString(string, MIME_TYPE.XML_TEXT).documentElement;
[...]

PR #895, opened 2025-07-07
fix compile error

Yancey2023:
**Public API Changes**

The `xml` parameter in `UrdfModelOptions` changed from type `Element | undefined` to `Element | null`

**Description**

fix a compile error in new version of xmldom:

```
src/urdf/UrdfModel.js:35:7 - error TS2322: Type 'Element | null' is not assignable to type 'Element | null | undefined'.
  Type 'Element' is missing the following properties from type 'Element': classList, className, clientHeight, clientLeft, and 103 more.

35       xmlDoc = parser.parseFromString(string, MIME_TYPE.XML_TEXT).documentElement;
```


Yancey2023:
To reproduce this compile error, clone this repo and run `npm install`.

MatthijsBurgh (changes requested):
Your change breaks the CI

Yancey2023:
Sorry about that. I only tested locally. I will try to fix it.

Yancey2023:
@MatthijsBurgh fixed

MatthijsBurgh:
@Yancey2023 Also make sure you update the lock file

Yancey2023:
Thanks for your review. I have update the lock file.

EzraBrooks:
Ah, did we never add the typescript compiler to CI here? I thought we had. We should probably add that to prevent dependabot version bumps from breaking the build.

EzraBrooks:
Actually, I see that the Vite build is running in CI here.
https://github.com/RobotWebTools/roslibjs/blob/2af39fa2b4b95b8db1929140227a248eec141966/test/build.bash#L28

In what situation are you encountering this build issue? I am not seeing an issue in CI or on my local machine.

Yancey2023:
The issue arises because the dependency versions in `package.json` are prefixed with a caret (^), which means any version above this is acceptable. As a result, when my computer has newer versions of dependencies, it uses those updated dependencies. However, these newer versions may be incompatible with 'roslib', which can lead to compilation errors.

Yancey2023:
Initially, I wasn't know the reason. Therefore, the PR initially only fixed the compilation errors without updating the dependency versions. Since the CI's npm uses caching, it didn’t use the newest dependency versions, causing the CI to initially fail.

Yancey2023:
I think others have also encountered this issue: https://github.com/RobotWebTools/roslibjs/issues/838#issuecomment-2607132861

Yancey2023:
There is also a PR that attempts to fix this issue, but it does so in an wrong way: https://github.com/RobotWebTools/roslibjs/pull/870

MatthijsBurgh (changes requested):
I am not sure whether we need all dependency bumps. Only bump dependencies of which the older version is not working anymore

Yancey2023:
I'm not sure it is a good idea because `npm` will try to install the newest version as default when version is prefixed with a caret (^) and the depencedy is not cached.

Yancey2023:
Hello. I can't understand why we're not using the newest version. Can you show me more details?

MatthijsBurgh:
Dependencies tell what you **need**, the lock file tells what you **use**.
So if we don't need a newer version, don't bump it. Improves compatibility for dependents

Yancey2023:
Sorry for the late reply and my misunderstanding. Everything is done now and I hope you can feel free to review it.

MatthijsBurgh (approved):

MatthijsBurgh:
@Yancey2023 Please resolve the conflicts

Yancey2023:
The conflicts is resolved so you can review now. By the way, I just notice that roslibjs was migrated to typescript and I think it's wonderful. I really love it.

package.json
```diff
@@ -70,4 +70,4 @@
   ],
   "types": "dist/RosLib.d.ts",
   "author": "Robot Webtools Team <robot-web-tools@googlegroups.com> (https://robotwebtools.github.io)"
-}
+}
```
MatthijsBurgh: Fix the missing newline at EOF

MatthijsBurgh (changes requested):

package.json
Yancey2023: got it

MatthijsBurgh (approved):

MatthijsBurgh:
@EzraBrooks @sea-bass Waiting for a 2nd approval from one of you before we should merge.

EzraBrooks (approved):
This seems right - I guess previously it was implicitly relying on the global-scoped `Element` type from JS itself, which is not necessarily equivalent to xmldom's `Element` type.

PR #903, opened 2025-08-05
Fix JSON encoding of double parameter values

ahoenerBE:
**Public API Changes**
Adds an optional parameter `numberTypeHint` to hint when parameter values should be encoded as doubles

**Description**
JSON stringification will cause a whole double value (e.g. 1.0) to be encoded as an int value ("1"). This will cause a rejection from the ros2 parameter service call unless dynamic typing is enabled, because it incorrectly interprets it as an int when the parameter type is Double.  
This PR adds a JSON replacer to the parameter encode to make sure that even if it's whole it has a decimal point.


EzraBrooks (commented):
Could you add a unit test? This is a great fix!

PR #905, opened 2025-08-08
Migrate library to TS (Base)

drewhoener:
**Public API Changes**
None

**Description**
This is a first pass at implementing changes that I discussed in my PR #851 

@EzraBrooks recommended a more incremental approach, and I'm finally back to deliver (I haven't had a lot of time because of project work, sorry).

### Overview
This PR is the start of my overall attempt to migrate RosLibJS to Typescript. While the library has generated types through JSDoc, it lacks the robustness that `TypeScript` and `typescript-eslint` can provide. Among other things, a move to TS can provide greater intellisense throughout the project (e.g. typed event names in EventEmitter methods, etc).

### Contents
I think the commits largely speak for themselves so I'll keep it brief, I tried to make the commit messages as detailed as I could for review.  
* Set up `tsconfig` with some more lenient rules for now, as the migration is incremental
* Fix the incorrect config def imported from `vite` instead of `vitest` and migrate away from a deprecated test config field
* Add type-aware eslint and typescript-eslint configs
* Rename index files to have a `.ts` extension to kickstart the process

#904 outlines my general plan for the refactor, I figured it would be good if I'm going to make multiple PRs

Please let me know if you have any feedback :)

Issue #904, opened 2025-08-08
Convert library to Typescript Roadmap


drewhoener:
This is an issue for tracking the migration of RosLibJS to Typescript. 

While the library has generated types through JSDoc, it lacks the robustness that `TypeScript` and `typescript-eslint` can provide. Among other things, a move to TS can provide greater intellisense throughout the project (e.g. typed event names in EventEmitter methods, etc).  

In terms of other motivation, I've been contributing to [RosReact](https://github.com/ltiseni94/rosreact) for the past several years, to the point that I've [forked it](https://github.com/boston-engineering/rosreact/tree/djh/build-refactor) since the original dev doesn't seem to have much time for it currently. I use it in almost every project that I design a dev interface for, it's a big part of my workflow now.  
I've ended up making a lot of type safe constructs to wrap around the RosLib objects and it would help my own development if the types I'm making on my end were reflected in this upstream library

### Base
- #909
- #905

### Migrate trivial modules
The following items can likely be migrated as-is without a ton of extra consideration for impact on the core modules, and likely have little to no extra refactoring work besides adding the types from docs.  

- [ ] #906
- [ ] #910
- [ ] #907

### Migrate Ros module & socket adapter
I have a good bit of this work completed from before I was requested to break the migration into parts, but it's a decent bit of refactoring of types.  
A possible breakdown:

- [ ] `cbor` array module in `utils` (#926)
- [ ] `SocketAdapter` (dependant on `cbor`) (#926)
- [ ] Add typings for EventEmitter events with their respective callback function types

### Migrate modules that use `Ros`
Basically just a catch-all for anything I didn't list above, but once the refactor to the `Ros` class is done, everything in these modules should fall into place fairly easily since it'll just be updating type information.  
This assumes that the overall API isn't changing much other than some generics or typing changes that need checking (which should be the case, I have some work done to support this as well since I started refactoring the above already). 

- [ ] Topic
- [ ] Message
- [ ] Param
- [ ] Service
- [ ] Action (ROS2)
- [ ] Actionlib (ROS1)
- [ ] `tf` module

### Tests
My original PR #851 has most of the work for migrating tests already done, so that shouldn't be too much of an issue, not going to bother with a checklist, just noting that they need to be updated

drewhoener:
@EzraBrooks whenever you have some free time, could you explain this to me?  

https://github.com/RobotWebTools/roslibjs/blob/develop/src/core/SocketAdapter.js#L138 
```js
[...]

PR #906, opened 2025-08-08
Migrate `math` module to TypeScript

drewhoener:
**Public API Changes**
None

**Description**

### Overview
This PR migrates the classes in the `math` module to Typescript. As I mentioned in #904, this PR is largely trivial and just requires adding the types described in the JSDoc.

### Contents
* Adds an interface `I<Type>` to each so math modules can reference each other without introducing circular imports.
* Use the same interfaces to provide constructor options
* Adds some utility types for null and undefined

Depends on #905 

GitHub doesn't have "dependent PRs" so all previous commits are included here. Might not be worth reviewing until dependencies are merged


EzraBrooks (commented):
GitHub tip: if you made this PR be based on your prior PR, it would deduplicate the diff - and when that PR is merged it would automatically retarget this one to the new base branch.

drewhoener:
> GitHub tip: if you made this PR be based on your prior PR, it would deduplicate the diff - and when that PR is merged it would automatically retarget this one to the new base branch.

Yeah, the 3 PRs I made are in a direct chain with each other, the branches continue from the previous one. I just mean github still shows all the commits until the previous one is merged. You can absolutely start reviewing at the first relevant commit though

drewhoener:
@EzraBrooks this one should also be ready

src/types/interface-types.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Represents a type that can either be a value of type `T` or `null`.
+ */
+export type Nullable<T> = T | null;
+
+/**
+ * Represents an object type where all properties are optional, and included properties can be null or undefined.
+ */
+export type PartialNullable<T> = {
+  [P in keyof T]?: Nullable<T[P]>;
+};
```
EzraBrooks: missing newline at EOF, can we enable https://eslint.style/rules/eol-last#eol-last ?

EzraBrooks (commented):
In modern typescript do we actually need the interfaces here or can we just import the class type as a type? I thought we could `import type MyClass` now 🤔

drewhoener:
> In modern typescript do we actually need the interfaces here or can we just import the class type as a type? I thought we could `import type MyClass` now 🤔

We can, the constructors define object types for their options though, so it made sense to make a type for it to also inherit.  Either that or I'd have to define a type anyway for the constructor since I'm trying not to change any API semantics.  
It also helps with DX for methods like `Quaternion.multiply`, if it accepts the interface type you don't have to be so verbose by making a new class instance.

```ts
const q: Quaternion = new Quaternion({
  x: 1,
  y: 1,
  z: 1,
  w: 1
});

// Without the interface type
q.multiply(new Quaternion({
  x: 1,
  y: 1,
  z: 1,
  w: 1
}));

// With type interface
q.multiply({
  x: 1,
  y: 1,
  z: 1,
  w: 1
});
```

You can still pass in a Quaternion instance, but it provides more flexibility

src/types/interface-types.ts
drewhoener: Fixed and added

EzraBrooks (approved):
Makes sense to me. I guess I'm just out of the object-oriented mindset 😆

PR #907, opened 2025-08-08
Migrate `urdf` modules to Typescript

drewhoener:
**Public API Changes**
None

**Description**

### Overview
This PR migrates the classes in the `urdf` module to Typescript. This PR is slightly more involved as there's a lot of interdependency between the URDF elements, but other than that it's still largely trivial.

### Contents
* Add TS enum for Urdf Type
* Add an enum to reduce magic strings when accessing attributes
* Extract logic for getting element origin into a separate function in `UrdfUtils.ts`
* Add stricter type checking for getting attributes from XML
* Use for-of loops added in ES6, building off my base PR. It's hit the "widely supported" threshold of 95% so there's really no reason not to use it.

Going to leave converting tests for another PR, but they all pass in my docker container

Depends on #906

GitHub doesn't have "dependent PRs" so all previous commits are included here. Might not be worth reviewing until dependencies are merged

### DIFF
--- a/src/urdf/UrdfColor.ts
+++ b/src/urdf/UrdfColor.ts
@@ -4,22 +4,41 @@
  * @author Russell Toris - rctoris@wpi.edu
  */
 
+import { UrdfAttrs, type UrdfDefaultOptions } from './UrdfTypes.js';
+import type { Optional } from '../types/interface-types.js';
+
 /**
  * A Color element in a URDF.
  */
 export default class UrdfColor {
+
   /**
-   * @param {Object} options
-   * @param {Element} options.xml - The XML element to parse.
+   * Color Red, [0, 1]
    */
-  constructor(options) {
+  r = 0.0;
+  /**
+   * Color Green, [0, 1]
+   */
+  g = 0.0;
+  /**
+   * Color Blue, [0, 1]
+   */
+  b = 0.0;
+  /**
+   * Alpha/Opacity, [0, 1]
+   */
+  a = 1.0;
+
+  constructor({ xml }: UrdfDefaultOptions) {
     // Parse the xml string
-    var rgba = options.xml.getAttribute('rgba')?.split(' ');
-    if (rgba) {
-      this.r = parseFloat(rgba[0]);
-      this.g = parseFloat(rgba[1]);
-      this.b = parseFloat(rgba[2]);
-      this.a = parseFloat(rgba[3]);
+    const rgba: Optional<string[]> = xml.getAttribute(UrdfAttrs.Rgba)?.split(' ');
+    if (!rgba || rgba.length !== 4) {
+      return;
     }
+
+    this.r = parseFloat(rgba[0]);
+    this.g = parseFloat(rgba[1]);
+    this.b = parseFloat(rgba[2]);
+    this.a = parseFloat(rgba[3]);
   }
 }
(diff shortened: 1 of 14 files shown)


EzraBrooks (commented):
GitHub is being GitHub here and the commit where you used `git mv` to ensure history continuity is not being represented in the PR diff. We might need to break that into its own PR (knowing it will probably break the build).

drewhoener:
> GitHub is being GitHub here and the commit where you used `git mv` to ensure history continuity is not being represented in the PR diff. We might need to break that into its own PR (knowing it will probably break the build).

GitHub's diff viewer might have a different threshold for what it considers a different file? At least locally with git in my terminal I can see the file history which is what I was hoping to preserve
image

EzraBrooks:
Yeah, unfortunately it looks like if I squash-merge this it will lose the history :(

drewhoener:
> Yeah, unfortunately it looks like if I squash-merge this it will lose the history :(

Created at #910

src/urdf/UrdfColor.ts
```diff
@@ -4,22 +4,41 @@
...
     // Parse the xml string
-    var rgba = options.xml.getAttribute('rgba')?.split(' ');
-    if (rgba) {
-      this.r = parseFloat(rgba[0]);
-      this.g = parseFloat(rgba[1]);
-      this.b = parseFloat(rgba[2]);
-      this.a = parseFloat(rgba[3]);
+    const rgba: Optional<string[]> = xml.getAttribute(UrdfAttrs.Rgba)?.split(' ');
+    if (!rgba || rgba.length !== 4) {
+      return;
```
EzraBrooks, line 36: Refactoring JS into TS always uncovers extremely odd control flow like this.. early return from a constructor, lol 🤦🏻

src/urdf/UrdfTypes.ts
```diff
@@ -1,4 +1,38 @@
...
+  Length = 'length',
+  Radius = 'radius',
+  Visuals = 'visual',
+  Texture = 'texture',
+  Filename = 'filename',
+  Color = 'color',
+  Geometry = 'geometry',
+  Material = 'material',
+  Scale = 'scale',
+}
```
EzraBrooks, line 31: This smells odd to me - I don't feel like we should need this enum to keep track of the magic strings (because each string would only be used once), but I understand you're just refactoring the code in-place here.

The URDF stuff could probably use a refactor soon. I've been using https://github.com/gkjohnson/urdf-loaders for years because I found the URDF implementation here roslibjs and in ros3djs too janky.

Not in the scope of this PR anyhow!

EzraBrooks: (also honestly maybe roslibjs and ros3djs should be merged into one package now that most of the world uses bundlers in their javascript apps and can tree-shake out unnecessary dependencies)

EzraBrooks (approved):
LGTM. Thanks for putting the effort in to doing this incrementally so I can scan it more easily!

drewhoener:
> LGTM. Thanks for putting the effort in to doing this incrementally so I can scan it more easily!

Yeah of course! Glad I finally have the time to get it done, I've been wanting to for a while. I feel like it's needed a bit of modernization.  
I probably won't have the next one until next week sometime, I need to think out the cbor/event emitter shenanigans and how that's all going to work.  

Thanks for being so responsive with my wave of PRs lol

EzraBrooks:
Yeah, I kept meaning to come back and do this refactor myself after I added Vite and tsc but never had time and at least in my application the JSDoc-generated types have been "fine" in the meantime

PR #908, opened 2025-08-08
(CI) Bump newest node version to 22

MatthijsBurgh:
**Public API Changes**
None

**Description**
Start using node 22

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           cache: npm
-          node-version: 20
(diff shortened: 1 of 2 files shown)


EzraBrooks (approved):

EzraBrooks:
We should probably also up the minimum engine version listed in the `package.json`. It hasn't been changed in years.

MatthijsBurgh:
@EzraBrooks do we have any reason to limit the engine version besides the versions we test and the versions that are EOL?

EzraBrooks:
I think we should just set the minimum engine version to whatever the minimum version we run in our tests is, that would make sense to me

PR #909, opened 2025-08-08
Refactor/typescript migration deps

drewhoener:
**Public API Changes**
None

### Overview
Prerequisite PR to #905, sets up `tsconfig.json` for a gradual conversion from JS -> TS.  

### Contents
* Change typescript version to match what's actually in package-lock.json
* Add eslint/js and typescript-eslint packages.
* Set up eslint to do strict type checking for all new TS code.
* Add a ts suppression to maintain compatibility


EzraBrooks (approved):
LGTM as long as it passes CI

PR #910, opened 2025-08-08
refactor(urdf): Rename files to TS.

drewhoener:
This commit will have lint/compile errors but is necessary for git continuity. Without it, git will think the files have been deleted and recreated because of the diff.

**Public API Changes**
None

**Description**
Rename URDF modules to TS to preserve file history in git


EzraBrooks (approved):
I'll see if I can circumvent the CI checks here since we'll fix them immediately after this

EzraBrooks:
Looks like no - we could temporarily comment out `vite-plugin-checker` in the vite config

drewhoener:
I can also make a commit to disable linting and then revert it in the next PR if we want to be strict about it

drewhoener:
> Looks like no - we could temporarily comment out `vite-plugin-checker` in the vite config

@EzraBrooks all set

PR #919, opened 2025-08-12
  Fix Service advertisement race conditions

### LINKED ISSUE #299: Service.advertise(callback) appends callbacks to the event listener for given service
If the same service is advertised, unadvertised, then advertised again, both callbacks given to service.advertise(callback) end up getting executed.  They keep stacking up the more you readvertise.

Example:
```js
    this.startUserConfirmationService = () => {
        this.userConfirmationService = new ROSLIB.Service({
            ros : ros,
            name : name,
            serviceType: type,
        })
    };
    this.startUserConfirmationService();

    this.advertiseUserConfirmation = () => {
        if (allowedToPublish) {
            this.userConfirmationService.advertise((req,res) => {
                if (confirm(req.info)) {
                    res.confirmation_response = true;
                    return true
                } else {
                    res.confirmation_response = false;
                    return true
                }
            })
        } else {
            this.userConfirmationService.unadvertise()
            console.log('alert! not allowed to send')
        }
    };

this.advertiseUserConfirmation() //allowedToPublish = true
// At this point, when the service is called there is 1 'confirm' dialog box
this.advertiseUserConfirmation() //allowedToPublish = false
// At this point, when the service is called, it is not available
this.advertiseUserConfirmation() //allowedToPublish = true
// At this point, when the service is called a 'confirm' dialog box pops up, followed by a second 'confirm' pup up

```

Adding ```this.ros.removeAllListeners(this.name) ``` before line 85 of src/core/Service.js: ``` this.ros.on(this.name, this._serviceResponse.bind(this));``` should solve this issue fairly simply.  I tested and it solved the issue for me.

noah-wardlow:
**Public API Changes**

  - `Service.advertise()` now returns a
  Promise that resolves when the
  advertisement is complete
  - `Service.advertiseAsync()` now returns a
  Promise that resolves when the
  advertisement is complete
  - `Service.unadvertise()` now returns a
  Promise that resolves when the
  unadvertisement is complete
  - Service errors are now emitted as
  'error' events instead of being logged to
  console

**Description**

 - Implement operation queue: All advertise/unadvertise
  operations are now serialized through a promise queue
  to prevent race conditions
  - Allow safe re-advertisement: advertise() and
  advertiseAsync() automatically unadvertise before
  re-advertising when a service is already advertised
  - Make operations async: Both methods now return
  promises that resolve when the operation completes
  - Prevent double operations: Added internal state
  tracking to prevent duplicate unadvertise operations
  - Fix cleanup order: Event callbacks are removed before
   sending unadvertise messages to stop processing new
  requests immediately

### DIFF
--- a/src/core/Service.js
+++ b/src/core/Service.js
@@ -19,6 +19,16 @@ export default class Service extends EventEmitter {
   _serviceCallback = null;
   isAdvertised = false;
   /**
+   * Queue for serializing advertise/unadvertise operations to prevent race conditions
+   * @private
+   */
+  _operationQueue = Promise.resolve();
+  /**
+   * Track if an unadvertise operation is pending to prevent double operations
+   * @private
+   */
+  _pendingUnadvertise = false;
+  /**
    * @param {Object} options
    * @param {Ros} options.ros - The ROSLIB.Ros connection handle.
    * @param {string} options.name - The service name, like '/add_two_ints'.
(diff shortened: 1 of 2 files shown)


EzraBrooks (commented):
Code LGTM - waiting for confirmation that it fixes the observed bug(s) in MoveIt Pro :)

EzraBrooks (approved):
Appears to solve the issues we're seeing with using services in component lifecycle in React!

PR #930, opened 2025-08-29
Handle rosapi param responses

bjsowa:
**Public API Changes**
None

**Description**
Thanks to RobotWebTools/rosbridge_suite#1001, since ROS 2 Jazzy, rosapi param services return the `successful` bool and fill the reason for failure if the operation does not succeed. 

This PR adds the logic for handling these additional fields.

### DIFF
--- a/src/core/Param.js
+++ b/src/core/Param.js
@@ -31,7 +31,7 @@ export default class Param {
    * Fetch the value of the param.
    *
    * @param {getCallback} callback - The callback function.
-   * @param {getFailedCallback} [failedCallback] - The callback function when the service call failed.
+   * @param {getFailedCallback} [failedCallback] - The callback function when the service call failed or the parameter retrieval was unsuccessful.
    */
   get(callback, failedCallback) {
     var paramClient = new Service({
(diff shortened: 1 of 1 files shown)


EzraBrooks:
was that backported to humble? roslibjs only has one release channel so I don't think we should break API here if not until Humble is EOL.

bjsowa:
> was that backported to humble? roslibjs only has one release channel so I don't think we should break API here if not until Humble is EOL.

It was not backported but this should still be compatible with humble

EzraBrooks (commented):
agree with copilot about not duplicating the code three places, otherwise LGTM. I only skimmed it the first time and didn't notice that you wrote it in a back-compatible way by checking for the presence of the `successful` member.

maybe we need to make `failedCallback` mandatory in a future version, otherwise if someone promisifies this callback-style function this'll just hang forever?

bjsowa:
@EzraBrooks I simplified it a bit by skipping the "undefined check" (`result.successful === false` will return false if `successful` is undefined) and also changed it so that the normal callback is still called if result is unsuccessful and failedCallback is not provided to keep the old behavior.  Let me know what you think

EzraBrooks (approved):

bjsowa:
@EzraBrooks Can we merge this?

PR #940, opened 2025-09-21
docs: add Quick Start Example to READMEAdded quick start example

FeliciaSeq:
This PR adds a Quick Start Example to the README, showing how to connect to a ROS master, handle connection events, and publish to a topic using roslibjs.

The goal is to help new developers quickly get started without needing to dig into examples or external docs.


MatthijsBurgh (changes requested):
I think it is good to point out the examples folder in the README. Though adding the code here, which is exactly the same code as in the simple node example, requires it to be maintained. Which I don't like.

PR #947, opened 2025-09-30
Drop node 18, add node 24

MatthijsBurgh:


sea-bass (approved):
LGTM, though you'll need to change the repo settings to make the node 18 test not required and the 24 one required

EDIT: Also seems all the checks are being duplicated by push and pull request events

MatthijsBurgh:
Yes, we run both events. Just to be sure the merge will be successful.

PR #973, opened 2025-11-03
feat: add fragment support for large messages sent by rosbridge

### LINKED ISSUE #557: Support rosbridge Fragmentation
The [rosbridge protocol provides the ability to fragment messages](https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/ROSBRIDGE_PROTOCOL.md#311-fragmentation--fragment--experimental).  This feature doesn't appear to be present in roslibjs.

Is there any plans to implement this? If not, is there some direction that can be provided in terms of getting this implemented?

For full context, I'm interested in using fragmentation in Foxglove, but [they explained](https://github.com/foxglove/studio/issues/4005) that they use roslibjs to handle this so such an update would have to occur here.

harsh-pal-robotics:
**Public API Changes**

None

**Description**

Messages sent to the websocket with the op fragment are not dealt with.
This buffers the incoming fragment messages and reuses the handleMessage function.

I had a service all where the result callbacks were never triggered because the event registered with the service call id was never triggered.

### DIFF
--- a/src/core/SocketAdapter.js
+++ b/src/core/SocketAdapter.js
@@ -30,7 +30,17 @@ export default function SocketAdapter(client) {
     decoder = client.transportOptions.decoder;
   }
 
+  /**
+   * Buffer Map for incoming message fragments
+   * @type {Map<string, {fragments: Array<string>, received: number, total: number}>}
+   */
+  const fragmentBuffer = new Map();
+
   function handleMessage(message) {
+    if (message.op === 'fragment') {
+      handleFragment(message);
+      return;
+    }
     if (message.op === 'publish') {
       client.emit(message.topic, message.msg);
     } else if (message.op === 'service_response') {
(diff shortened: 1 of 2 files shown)


EzraBrooks (approved):
Sweet!! This was on my to-do list long ago but once I worked around it by disabling fragmentation on the bridge side I forgot about it, honestly. The logic looks correct to me and the test coverage is awesome.

src/core/SocketAdapter.js
```diff
@@ -54,6 +64,46 @@ export default function SocketAdapter(client) {
...
+    // If total is a float, use its integer part for fragment count
+    const totalInt = Math.floor(total);
+    if (!fragmentBuffer.has(id)) {
+      fragmentBuffer.set(id, { fragments: [], received: 0, total: totalInt });
+    }
+    const entry = fragmentBuffer.get(id);
+
+    if (!entry) {
+      // Should not happen, keeping TypeScript happy
+      return;
```
EzraBrooks: The one minor non-blocking comment I have here is that this should probably throw an exception. I think we all know that things that should never happen tend to happen 😉

PR #976, opened 2025-11-03
Fix CI to use locked dependencies

EzraBrooks:
`npm ci` installs from package-lock instead of package,
so it's idempotent and also faster.


EzraBrooks:
superseded by #984

PR #977, opened 2025-11-03
Lint in CI

EzraBrooks:
I thought vite-plugin-checker was going to handle this for us, but it seems like some things have fallen through the cracks.


EzraBrooks:
or.. not? why is my editor locally upset about things that this isn't catching?

PR #978, opened 2025-11-03
Bump version in package.json to 2.0.0

EzraBrooks:
whoops, we should've done this quite awhile ago.


sea-bass (approved):

sea-bass:
To be fair, 2.0.0 hasn't yet been released, so...

bjsowa (approved):

EzraBrooks:
@sea-bass yeah but building this package locally it looks like 1.4.1 which it is very much not!

bjsowa:
Is there a way to use something like `git describe --tags` at build time?

PR #980, opened 2025-11-03
Improve PNG decompression implementation

### LINKED ISSUE #859: Request to make roslibjs usable in web workers
In SocketAdapter.js, I suggest using fast-png, a pure javascript library, to make shim/decompressPng.js usable in web workers

https://github.com/image-js/fast-png

EzraBrooks:
Use fast-png to unify the implementation between Node.js
and browser, and unbork Web Worker support (fixes #859)


EzraBrooks:
Hey @tosscaster could you test this? I don't have a Web Worker environment to test this in.

Or even helping add an automated test would be awesome.

EzraBrooks:
Given that there's no integration test, I think we should YOLO this, unless @bjsowa @sea-bass you guys have any reference for this w/r/t how it works in rosbridge_suite. I would hope that switching from one PNG implementation to another wouldn't break anything...

EzraBrooks:
okay, added a test. now we actually know for sure that PNG compression works!

EzraBrooks:
except for the fact that it times out in CI because it's slow as hell >:(

EzraBrooks:
The test timeout isn't from the roslibjs side of things. The initial run before the V8 JIT optimizes it only takes 3.7ms, and then subsequent runs take under half a millisecond.

```
Decompressing PNG: 3.773ms
Decompressing PNG: 0.223ms
Decompressing PNG: 0.154ms
Decompressing PNG: 0.159ms
Decompressing PNG: 0.319ms
```

EzraBrooks:
Interestingly, on my most recent attempt before lengthening the timeout, all the ROS 2 jobs failed but the ROS 1 jobs succeeded.

EzraBrooks:
@bjsowa @sea-bass it looks like there's something wrong with rosbridge here, in ROS 2 specifically.

bjsowa:
I'll try to take a look later today

bjsowa:
@EzraBrooks I see what's the problem. There was a bug with PNG compression in rosbridge which was fixed some time ago (RobotWebTools/rosbridge_suite#1045). The fix is included in the newest versions of rosbridge which I release a week ago.

I tried the example against the newest version of rosbridge in Jazzy and Rolling and it worked.

EzraBrooks:
So does the introduction of this test need to wait til the next sync?

bjsowa:
> So does the introduction of this test need to wait til the next sync?

Unless we change the dockerfile to use ROS testing repository or build rosbridge itself.

EzraBrooks:
Given that the code works, I'm inclined to mark the test skipped and merge this. Are you okay with that?

EzraBrooks:
Note to self: remove pngparse type declaration file here before merging because the dependency itself is gone

bjsowa:
> Given that the code works, I'm inclined to mark the test skipped and merge this. Are you okay with that?

I'm fine with that

bjsowa (approved):

bjsowa:
There should be a new sync for each ros distro by the end of this month. After the sync and when the docker images are updated, we should be able to reenable the test.

PR #979, opened 2025-11-03
align indentation in rest of lib with core

EzraBrooks:
The `src/core` directory was two-space indentation (my preference also) but seemingly nothing else was. This PR fixes that and codifies the style in the ESLint config.

PR #981, opened 2025-11-03
Enable `noImplicitAny` and fix violations

EzraBrooks:
Pretty naive fixing for now, just trying to get the codebase to a point where the data types are traceable throughout.

Merge #979 first, I based this on that to avoid massive conflicts.

### DIFF
--- a/src/types/pngparse.d.ts
+++ b/src/types/pngparse.d.ts
@@ -1,0 +1,22 @@
+declare module 'pngparse' {
+  interface ParsedImage {
+    width: number;
+    height: number;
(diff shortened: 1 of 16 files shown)


src/types/pngparse.d.ts
EzraBrooks, line 1: This can be deleted if #980 gets merged first.

EzraBrooks:
actually.. I think I should do this the other way around first. Disable _all_ the `any` checks, rename all the `.js` files to `.ts`, and then incrementally re-enable those checks.

PR #984, opened 2025-11-03
Move integration test env into container, launch in Vitest

EzraBrooks:
Restructures the testing environment for roslibjs to automatically build & launch a Docker container containing the ROS backend code while running the JavaScript frontend code on the host.


EzraBrooks:
My vision here is that we can get this to a point where the test environment can be spun up during setup of the Vitest run and spun down during cleanup of the Vitest run, so contributors can easily run integration tests without setting up their computer with the exact version(s) of ROS we target.

package-lock.json
EzraBrooks, line 1: Most of the diff is just adding `dockerode` as a dev dependency adding 700 lines to the lock file!

package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" ?>
+<package>
+  <name>roslibjs</name>
+  <version>1.4.1</version>
```
sea-bass, line 4: make sure this is bumped to 2.0.0 in the other PR that bumps versions, maybe land this first?

sea-bass (approved):

PR #985, opened 2025-11-04
feat: expose rotation axis info in UrdfJoint

harsh-pal-robotics:
**Public API Changes**

None

**Description**

On some of the urdf joints the default axis is not X, this changes exposes the information adding an `axis` field to the URDF Joint.
by default if the value is not available or parseable, use X as fallback.


src/urdf/UrdfJoint.ts
```diff
@@ -47,5 +51,18 @@ export default class UrdfJoint {
     if (origins.length > 0) {
       this.origin = parseUrdfOrigin(origins[0]);
     }
+
+    const axis = xml.getElementsByTagName(UrdfAttrs.Axis);
+    if (axis.length > 0) {
+      const xyzValue = axis[0].getAttribute(UrdfAttrs.Xyz)?.split(' ');
+      if (xyzValue && xyzValue.length === 3) {
```
EzraBrooks: The `axis` element itself is optional, however according to the URDF spec the `xyz` attribute is mandatory if `axis` exists.

I think we should fail hard (throw) if xyzValue is null or malformed.

EzraBrooks (commented):
Great enhancement! one comment about possibly validating the schema

src/urdf/UrdfJoint.ts
harsh-pal-robotics: okay, make sense!

EzraBrooks (approved):
Sorry to hijack your PR :wink:. You did most of the work, I just wanted to get it in!

PR #986, opened 2025-11-04
feat: add functions to retrieve ROS action details

harsh-pal-robotics:
**Public API Changes**

ROS class has new methods to retrieves information of action interfaces
- getActionGoalDetails
- getActionResultDetails
- getActionFeedbackDetails

**Description**

Added functions to retrieve the action goal, result, and feedback details.
Implementation in rosbridge: 

- https://github.com/RobotWebTools/rosbridge_suite/pull/1021
- https://github.com/RobotWebTools/rosbridge_suite/pull/1065

### DIFF
--- a/src/core/Ros.js
+++ b/src/core/Ros.js
@@ -708,6 +708,121 @@ export default class Ros extends EventEmitter {
     }
   }
   /**
+   * @callback getActionGoalDetailsCallback
+   * @param {{typedefs: string[]}} result - The result object with the following params:
+   */
+  /**
+   * @callback getActionGoalDetailsFailedCallback
+   * @param {string} error - The error message reported by ROS.
+   */
+  /**
+   * Retrieve the details of a ROS action goal.
+   *
+   * @param {string} type - The type of the action.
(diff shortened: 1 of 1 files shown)


EzraBrooks (commented):
Given that these are some fairly new APIs in rosapi, should we check the existence of the service before calling it, and tell the user their rosapi version might be too old?

harsh-pal-robotics:
> Given that these are some fairly new APIs in rosapi, should we check the existence of the service before calling it, and tell the user their rosapi version might be too old?

hmmm not sure if its needed to throw an error, it will already trigger the error callback with the service does not exist message.

But if we want to throw a message, we need to create a function first to check if a service available, and a ServiceNotAvailable error can be created and thrown.

EzraBrooks (commented):
Okay, I think it's fine to not add a check.

However, in the time since you opened this PR, I started adding types for the rosapi message types into the typescript portion of the codebase - could you add these message definitions there, to use in the JSDoc comments you added? Thanks!

harsh-pal-robotics:
closing this for this one in favor of https://github.com/RobotWebTools/roslibjs/pull/1213

PR #989, opened 2025-11-04
Remove misleading React example

EzraBrooks:
The React example code is actually quite misleading and displays
major anti-patterns (such as constructing an un-memoized Ros instance
in component render code) so it is better to remove it to stop confusing
users.


EzraBrooks:
imo we should point users in the direction of [react-ros](https://github.com/flynneva/react-ros) or similar until we can provide a first-party `roslibreact` wrapper of `roslibjs`.

Which is something I absolutely have the know-how to do quite fast, once we've shipped roslibjs 2.0!

sea-bass (approved):
this is a good plan

PR #991, opened 2025-11-04
Further modernize class definitions

EzraBrooks:
Two major cleanups:

- Destructure options objects in the class constructor signature to make defaults clearer
- Add missing class member declarations for members that are assigned in the constructor but not declared. This will be required when moving to TypeScript and has made the type inference better when working in the JS code in the meantime.

## This is not a breaking change, it just changes some syntactic sugar and explicitly declares some things that were previously implicitly declared!

### DIFF
--- a/src/actionlib/ActionClient.js
+++ b/src/actionlib/ActionClient.js
@@ -32,15 +32,23 @@ export default class ActionClient extends EventEmitter {
    * @param {boolean} [options.omitStatus] - The flag to indicate whether to omit the status channel or not.
    * @param {boolean} [options.omitResult] - The flag to indicate whether to omit the result channel or not.
    */
-  constructor(options) {
+  constructor({
+    ros,
+    serverName,
+    actionName,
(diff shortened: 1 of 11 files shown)


sea-bass (commented):
Looks good, but it's a breaking change, right? Did you intend this to go in before or after the 2.0.0 cutoff?

EzraBrooks:
I don't think there's any API break here, the signature of the constructors remains the same to customers, it's just destructuring it inline.

EzraBrooks:
diff has some whitespace noise in it, opened #995 to fix

EzraBrooks:
I plan on incrementally adding JSDoc type annotations to these members in future PRs, thus allowing these files to just be in-place renamed from `.js` to `.ts` and ✨ magically be TypeScript ✨.

bjsowa (approved):
There are still a few formatting changes but I don't see a problem with including them

PR #996, opened 2025-11-05
Align ESLint config between JS and TS

EzraBrooks:
This means (temporarily) cranking down the strictness of the TS linting, but it has cranked the JS strictness way up. Trying to get the two as close together as possible to ease the migration.


EzraBrooks:
After this PR and #991  it's now possible to rename things from `.js` -> `.ts` and have them work without further changes 🎉

bjsowa (approved):

PR #999, opened 2025-11-05
Add protocol type validation

EzraBrooks:
Added TypeScript interfaces for the whole rosbridge protocol (as documented in the markdown file in rosbridge - maybe we should make a proper JSON schema that can be interpreted into types in our various languages), added type guards that mark an object as the right type dependent on its `op` field, and used those type guards in the (most important) places where messages are directly interacted with.

### DIFF
--- a/src/core/Action.js
+++ b/src/core/Action.js
@@ -5,6 +5,7 @@
 
 import { EventEmitter } from 'eventemitter3';
 import { GoalStatus } from '../core/GoalStatus.ts';
+import { isRosbridgeActionFeedbackMessage, isRosbridgeActionResultMessage, isRosbridgeCancelActionGoalMessage } from '../types/protocol.ts';
 
 /**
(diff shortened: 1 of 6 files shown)


bjsowa (commented):
Looks good. I actually have something similar for Rosbridge library in my TODO list.

Just a quick question since I'm not proficient in typescript: Will this also check if the message has required fields?

EzraBrooks:
I did not add any validation of the individual fields within the messages yet, no. At the moment it's just doing the same thing it was doing before - checking `op` - but doing it in such a way that it informs the type system.

Going any further into validation I would probably stop writing the type guards by hand and instead adopt a framework like [zod](https://zod.dev) where you can define schemas and it generates type guards.

bjsowa (approved):
Got it! If you decide to write JSON schemas in the future, I'll be down to add better validation in Rosbridge library.

PR #1000, opened 2025-11-05
Refactor SocketAdapter from mixin pattern to a wrapper

EzraBrooks:
I think this is a more sensical approach to this problem. SocketAdapter provides a coherent ROS API on top of either a browser WebSocket, a ws WebSocket, or an RTCDataChannel - it should be written as a wrapper.

This technically breaks an API, but only because of Hyrum's Law.. I hope nobody was interacting directly with the `.socket` member of `Ros` 🤦🏻 

## This contains no functional changes, it just changes the scope of some of the socket-handling callbacks to be less confusing.

### DIFF
--- a/src/core/Ros.js
+++ b/src/core/Ros.js
@@ -4,6 +4,16 @@
  */
 
 import socketAdapter from "./SocketAdapter.js";
+import {
+  isRosbridgeActionFeedbackMessage,
+  isRosbridgeActionResultMessage,
+  isRosbridgeCallServiceMessage,
+  isRosbridgeCancelActionGoalMessage,
+  isRosbridgePublishMessage,
+  isRosbridgeSendActionGoalMessage,
+  isRosbridgeServiceResponseMessage,
+  isRosbridgeStatusMessage,
+} from "../types/protocol.js";
 
 import Topic from "./Topic.js";
 import Service from "./Service.js";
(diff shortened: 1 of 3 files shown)


bjsowa:
There are some conflicts

EzraBrooks:
just GitHub being weird about not being able to automatically resolve a renamed file. I don't understand why Git locally has no problem with that but GitHub does. fixed

src/core/Ros.js
```diff
@@ -4,6 +4,7 @@
  */
 
 import socketAdapter from './SocketAdapter.js';
+import { isRosbridgeActionFeedbackMessage, isRosbridgeActionResultMessage, isRosbridgeCallServiceMessage, isRosbridgeCancelActionGoalMessage, isRosbridgePublishMessage, isRosbridgeSendActionGoalMessage, isRosbridgeServiceResponseMessage, isRosbridgeStatusMessage } from '../types/protocol.js';
```
bjsowa: Don't we have some limit on line length? 294 characters seems a bit excessive...

EzraBrooks: apparently not! I think the style rules in this repo are a bit lacking.

I'll open a PR to just adopt the standard eslint "stylistic" config instead of continuing to limp along with the one we've had for >10yrs.

EzraBrooks: #1003

bjsowa (approved):

PR #1003, opened 2025-11-06
Replace our custom code style with Prettier

EzraBrooks:
Prettier has become the de facto code formatter for the web ecosystem. Let's just use it instead of continuing to home-brew our own style.

First commit is the config change, second commit is just a formatter run.


EzraBrooks:
I considered `eslint-plugin-stylistic` but frankly felt it didn't go far enough and we'd just revisit this decision again later if we adopted it.

bjsowa (approved):
+1 for using prettier. I quickly skimmed through the changes and didn't see anything malformed.

Issue #335, opened 2019-06-20
SimpleActionServer should expose dispose()


aez:
There is currently no way to tear down a SimpleActionServer instance. A dispose() method seems like a reasonable thing to add to the API.

Rayman:
Indeed, it should be relatively easy to implement. In the action client it's implemented here: https://github.com/RobotWebTools/roslibjs/blob/f0d6d6a6dcaadb8abacb47a5a82b126af79f84e8/src/actionlib/ActionClient.js#L135

PR #1005, opened 2025-11-06
Port SimpleActionServer to TypeScript, fix Action timestamping

EzraBrooks:
Did this one next because the right solution to #335 should be to make this class inherit from TypeScript's `Disposable` interface.

Also, funny thing about this - turns out the original code was never timestamping anything properly because it wasn't strictly typed. It had typos in the names of the time members 😆

### DIFF
--- a/src/types/actionlib_msgs.ts
+++ b/src/types/actionlib_msgs.ts
@@ -1,4 +1,4 @@
-import { std_msgs } from "../types/std_msgs";
+import { std_msgs } from "./std_msgs.js";
 import { GoalStatus as GoalStatusEnum } from "../core/GoalStatus.js";
 
 export namespace actionlib_msgs {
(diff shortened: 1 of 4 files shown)


src/types/actionlib_msgs.ts
bjsowa, line 1: Wouldn't it be more appropriate to move this to `src/types`?

src/types/std_msgs.ts
bjsowa, line 1: This type is slightly different in ROS 2. Should probably be at least mentioned.

bjsowa (changes requested):

src/types/std_msgs.ts
EzraBrooks: Renamed to `ROS1Header` since so far it's only used by the actionlib stuff which is ROS1 only

src/types/actionlib_msgs.ts
EzraBrooks: done

bjsowa (approved):

PR #1008, opened 2025-11-06
Port Service to TypeScript

EzraBrooks:


src/types/protocol.ts
```diff
@@ -142,12 +142,12 @@ export function isRosbridgeUnadvertiseServiceMessage(
   return message.op === "unadvertise_service";
 }
 
-export interface RosbridgeCallServiceMessage<TArgs = unknown[]>
+export interface RosbridgeCallServiceMessage<TArgs = void>
   extends RosbridgeMessage {
   op: "call_service";
   id?: string;
   service: string;
-  args?: TArgs;
```
bjsowa, line 150: Rosbridge protocol description says `args` field is optional

bjsowa: Also `values` in `RosbridgeServiceResponseMessage` can be of type `string` when the `result` is set to false

EzraBrooks: This change ensures that the field is only optional when the service itself takes no arguments, i.e. a Trigger service, which I think is the intent of the optionality in the protocol?

I made the field "non-optional" but made its type default to void, which effectively means if you don't specify a type it'll be left empty.

EzraBrooks: > Also `values` in `RosbridgeServiceResponseMessage` can be of type `string` when the `result` is set to false

That presents an interesting polymorphism issue.. I'll see if I can manage to address it without a major overhaul

bjsowa: AFAIK Rosbridge server allows `args` to be an Object, an Array or not provided at all, even when service takes arguments. When sending `call_service` messages to clients, it will always send an Object though.

EzraBrooks: I don't think roslibjs should allow its consumers to send unrelated data types, but maybe it should allow Partials. The logic in rosbridge will deeply fill missing fields with defaults, right? Maybe this should be some sort of `DeepPartial<TArgs>`

src/types/protocol.ts
```diff
@@ -159,15 +159,30 @@ export function isRosbridgeCallServiceMessage(
...
+
+interface SuccessfulRosbridgeServiceResponseMessage<TValues = undefined>
+  extends BaseRosbridgeServiceResponseMessage {
+  values: TValues;
+  result: true;
+}
+
+export type RosbridgeServiceResponseMessage<TValues = undefined> =
+  | FailedRosbridgeServiceResponseMessage
+  | SuccessfulRosbridgeServiceResponseMessage<TValues>;
```
EzraBrooks, line 188: @bjsowa I've created a union type for `RosbridgeServiceResponseMessage` that enforces that the value returned is of type `TValues` if successful but is of type `string` if unsuccessful, does that sound right?

bjsowa: That sounds about right

src/types/protocol.ts
bjsowa: Yep, rosbridge fills missing fields with defaults

bjsowa: As for passing Array as `args`. I don't think we need to support it in Typescript port.

EzraBrooks: I fiddled with making this a deep partial but unfortunately had no luck keeping the interface sane while doing so - since there's no distinction between an "incoming call service request" to a service server hosted by roslibjs, which _would not be partial_, and an "outgoing call service request", which might be partial. I think we should leave it be for now. I'll leave a TODO

src/core/Service.ts
```diff
@@ -91,20 +101,17 @@ export default class Service extends EventEmitter {
...
    * that's made on this service.
    *
-   * @param {advertiseCallback} callback - This works similarly to the callback for a C++ service and should take the following params
+   * @param callback This works similarly to the callback for a C++ service in that you should take care not to overwrite the response object.
+   *  Instead, only modify the values within.
    */
-  advertise(callback) {
+  advertise(
+    callback: (request: TRequest, response: Partial<TResponse>) => boolean,
+  ) {
```
bjsowa: Specify type for API clarity
```suggestion
  ): Promise<void> {
```

bjsowa: The same for other public methods

src/core/Service.ts
```diff
@@ -118,18 +125,22 @@ export default class Service extends EventEmitter {
           const response = {};
           const success = callback(rosbridgeRequest.args, response);
```
bjsowa: You wrap the callback in `try/catch` in `advertiseAsync` but not here

src/core/Service.ts
```diff
@@ -4,18 +4,23 @@
...
-export default class Service extends EventEmitter {
+export default class Service<TRequest, TResponse> extends EventEmitter {
   /**
    * Stores a reference to the most recent service callback advertised so it can be removed from the EventEmitter during un-advertisement
    * @private
-   * @type {((rosbridgeRequest) => any) | null}
    */
-  _serviceCallback = null;
+  _serviceCallback:
+    | ((rosbridgeRequest: RosbridgeCallServiceMessage<TRequest>) => void)
```
bjsowa: Shouldn't this return `void | Promise<void>` since in advertiseAsync you assign an async function?

src/core/Service.ts
EzraBrooks: good point - this is an existing issue, but worth adding while we're in here.

bjsowa (approved):

PR #1009, opened 2025-11-06
Port TFClient impls to TypeScript

EzraBrooks:
Lots of hand-written message types here.. need to figure out a way to automate `ros-typescript-generator` using the Docker container I built for the integration tests

### DIFF
--- a/src/tf/TFClient.ts
+++ b/src/tf/TFClient.ts
@@ -8,6 +8,9 @@ import Goal from "../actionlib/Goal.js";
 
 import Service from "../core/Service.js";
 import Topic from "../core/Topic.js";
+import Ros from "../core/Ros.js";
+import { tf2_msgs } from "../types/tf2_msgs.js";
+import { tf2_web_republisher } from "../types/tf2_web_republisher.js";
 
 import BaseTFClient from "./BaseTFClient.js";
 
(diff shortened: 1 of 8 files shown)


src/tf/TFClient.ts
```diff
@@ -8,35 +8,53 @@ import Goal from "../actionlib/Goal.js";
...
+    repubServiceName = "/republish_tfs",
+    ...options
+  }: {
+    ros: Ros;
+    fixedFrame?: string;
+    angularThres?: number;
+    transThres?: number;
+    rate?: number;
+    updateDelay?: number;
+    topicTimeout?: number;
```
sea-bass, line 54: preceded this PR, but `topicTimeout` should really be `serviceTimeout` ... however, not worth breaking anything for this i guess.

src/types/geometry_msgs.ts
```diff
@@ -0,0 +1,22 @@
+export namespace geometry_msgs {
+  export interface TransformStamped {
+    // TODO: some sort of unholy union type of ros1 header and ros2 header???
```
sea-bass, line 3: ros 1 is deprecated, so I'd just not bother and point people to use older versions of this library if they need ros 1

src/types/geometry_msgs.ts
```diff
@@ -0,0 +1,22 @@
...
+  export interface Vector3 {
+    x: number;
+    y: number;
+    z: number;
+  }
+  export interface Quaternion {
+    x: number;
+    y: number;
+    z: number;
+    w: number;
```
sea-bass, line 20: minor point, but in the ROS 2 message definition, the `w` field actually defaults to 1.0, whereas all the other scalar numeric fields default to 0.0. Is that something we can/should do here too?

src/types/tf2_web_republisher.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  export interface TFSubscriptionGoal {
+    source_frames: string[];
+    target_frame: string;
+    angular_thres: number;
+    trans_thres: number;
+    rate: number;
+  }
+
+  export type TFSubscriptionResult = Record<never, never>;
```
sea-bass, line 23: what does it mean that this one is `type` vs the others being `interface`?

src/tf/TFClient.ts
EzraBrooks: Preceded this PR by many years I think!!

src/types/geometry_msgs.ts
EzraBrooks: Agreed in theory, but at that point why not rip out the entire ROS 1 implementation of TFClient?

EzraBrooks: and actionlib..

src/types/geometry_msgs.ts
EzraBrooks: interfaces can't be instantiated, so you can't set defaults

src/types/tf2_web_republisher.ts
EzraBrooks: `type` lets me use type expressions, like the assignment I'm doing here. `interface` only allows describing an object type as a block of fields in curly braces

sea-bass (approved):
I think all my initial comments are eh, let's do it

PR #1014, opened 2025-11-07
Port ROS 1 Goal to TypeScript

EzraBrooks:
another simple one, and yet again with typos in the timestamp field names :rage2:

### DIFF
--- a/src/actionlib/Goal.ts
+++ b/src/actionlib/Goal.ts
@@ -4,6 +4,8 @@
  */
 
 import { EventEmitter } from "eventemitter3";
+import ActionClient from "./ActionClient";
+import { actionlib_msgs } from "../types/actionlib_msgs";
 
 /**
  * An actionlib goal that is associated with an action server.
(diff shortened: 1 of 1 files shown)


sea-bass (approved):

sea-bass:
so fun fact, it seems like the stamp fields changed from `secs` to `sec` between ros 1 and 2, because rosbridge seems to accept them both?

https://github.com/RobotWebTools/rosbridge_suite/blob/7426843a88b2bf4795fb653daa6e6a5e1f9cbf8b/rosbridge_library/src/rosbridge_library/internal/message_conversion.py#L422-L430

EzraBrooks:
what.. but [the documentation for `ros::time`](https://wiki.ros.org/roscpp/Overview/Time) has `sec` and `nsec`

EzraBrooks:
oh lord..

https://design.ros2.org/articles/changes.html

> In ROS 1 the duration and time types are defined in the client libraries. The member names of the data structures are different in C++ (sec, nsec) and Python (secs, nsecs).

well, good thing rosbridge rectifies them.

EzraBrooks:
I'm gonna 

in my next PR. I want roslibjs to be internally consistent even if that means using the ros1 stuff for now

PR #1015, opened 2025-11-07
Port ROS 1 ActionListener to TypeScript

EzraBrooks:


src/actionlib/ActionListener.ts
```diff
@@ -17,42 +19,60 @@ import { EventEmitter } from "eventemitter3";
...
-    const goalListener = new Topic({
+    const goalListener = new Topic<TGoal>({
       ros: this.ros,
       name: this.serverName + "/goal",
       messageType: this.actionName + "Goal",
     });
 
-    const feedbackListener = new Topic({
+    const feedbackListener = new Topic<{
+      status: actionlib_msgs.GoalStatus;
```
sea-bass, line 58: is `GoalStatus` means to be in the feedback listener?

EzraBrooks: I, too, am baffled by this. But the code does refer to it.

EzraBrooks: it really do be like this:

https://github.com/ros/actionlib/blob/c76c8e67c95ea05c8eb969ca4fa804c094d4a18a/actionlib/src/actionlib/server_goal_handle.py#L213

EzraBrooks: ooh there actually may be something more sinister going on in there.. this might be a _different_ "status"

EzraBrooks: nope nvm

https://github.com/ros/actionlib/blob/c76c8e67c95ea05c8eb969ca4fa804c094d4a18a/actionlib/src/actionlib/status_tracker.py#L49

I was right the first time. lmao.

sea-bass (approved):
well then, lol

PR #1016, opened 2025-11-07
Port ROS 1 ActionClient to TypeScript

EzraBrooks:


src/actionlib/Goal.ts
```diff
@@ -20,7 +20,7 @@ export default class Goal<T> extends EventEmitter {
   feedback = undefined;
   // Create a random ID
   goalID = "goal_" + Math.random() + "_" + new Date().getTime();
-  actionClient: ActionClient;
+  actionClient: ActionClient<T, unknown, unknown>;
```
EzraBrooks: looks a bit odd but the Goal doesn't need to know the result and feedback types.

sea-bass (approved):

PR #1018, opened 2025-11-07
Port Ros to TypeScript

EzraBrooks:
This is the big one.

## There is _one_ breaking API change in this PR! I removed the "legacy callback" from `getNodeDetails` which was a workaround from _9 years ago!_ I think we're good to break that for 2.0.


sea-bass (approved):
Good with removing that one legacy thing you mentioned 🚢

PR #1020, opened 2025-11-07
Port dev config files to TypeScript

EzraBrooks:
Why didn't I just write these in TypeScript in the first place...


package.json
```diff
@@ -27,6 +27,7 @@
     "eslint-plugin-jsdoc": "^61.1.12",
     "eslint-plugin-prettier": "^5.5.4",
     "globals": "^16.0.0",
+    "jiti": "^2.6.1",
```
EzraBrooks, line 30: needed for typescript eslint config

sea-bass (approved):
Why let a linter do things you could do yourself through sheer willpower (and free time)?

PR #1021, opened 2025-11-07
Enforce member privacy using Private Elements

### LINKED ISSUE #1010: After TypeScript migration, consolidate member privacy from `_` and `@private` to `#`
[Private elements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) finally became a thing in the underlying ECMAScript standard a few years back, so we can totally skip the awkward era of the TypeScript `private` keyword that was private at build time but public at run time.

Just go through and replace things documented as `@private` or prefixed with `_` with a `#` prefix.

EzraBrooks:
**Public API Changes**

Anyone who was relying on things marked "private" is gonna actually experience them being private now, which is technically an API break.

**Description**

Migrates the two ways members were marked as private in this repo - `@private` in JSDoc and a leading `_` - to the modern ECMAScript `#` that is enforced by both TypeScript at build-time and modern JS implementations at run-time.

Closes #1010

### DIFF
--- a/src/tf/BaseTFClient.ts
+++ b/src/tf/BaseTFClient.ts
@@ -12,7 +12,6 @@ export default class BaseTFClient extends EventEmitter {
     { transform?: Transform; cbs: ((tf: Transform) => void)[] }
   > = {};
   republisherUpdateRequested = false;
-  _isDisposed = false;
   ros: Ros;
   fixedFrame: string;
   angularThres: number;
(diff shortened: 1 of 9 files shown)


src/tf/BaseTFClient.ts
```diff
@@ -12,7 +12,6 @@ export default class BaseTFClient extends EventEmitter {
     { transform?: Transform; cbs: ((tf: Transform) => void)[] }
   > = {};
   republisherUpdateRequested = false;
-  _isDisposed = false;
```
EzraBrooks, line 15: This was erroneously in the shared base class when only used by one of the children.

sea-bass (approved):

PR #1023, opened 2025-11-07
Statically type EventEmitter events

EzraBrooks:
**Public API Changes**

none

**Description**

Routed out another few bits of weirdness by statically typing the event interfaces of each class. Funnily enough, this revealed that some of them were just not being used at all and could be removed.

### DIFF
--- a/src/core/Ros.ts
+++ b/src/core/Ros.ts
@@ -35,7 +35,14 @@ import { rosapi } from "../types/rosapi.ts";
  *  * &#60;topicName&#62; - A message came from rosbridge with the given topic name.
  *  * &#60;serviceID&#62; - A service response came from rosbridge with the given ID.
  */
-export default class Ros extends EventEmitter {
+export default class Ros extends EventEmitter<
+  {
+    error: [string];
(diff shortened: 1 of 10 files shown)


src/core/Ros.ts
```diff
@@ -148,7 +155,7 @@ export default class Ros extends EventEmitter {
    */
   handleMessage(message: RosbridgeMessage) {
     if (isRosbridgePublishMessage(message)) {
-      this.emit(message.topic, message.msg);
+      this.emit(message.topic, message);
```
sea-bass, line 156: Mostly curious about this one -- was it a bug?

sea-bass (approved):

src/core/Ros.ts
EzraBrooks: Not a bug as much as a weird layer of indirection - you can see in `Topic` that it was (internally, not exposed to users) behaving slightly differently than the other classes. It was handling just the inner `msg` member instead of the whole message from rosbridge. By moving the place where `.msg` was being accessed, the API became consistent.

PR #1024, opened 2025-11-08
Add ROS 2 integration tests

### LINKED ISSUE #746: Add ROS 2 integration tests
We'd like to release roslibjs v2 with a lot of improvements (asynchronous Services, ROS 2 Actions, ROS 2 TFClient), but we don't yet have any integration tests in CI for the ROS 2 support. Once this is done, we can release v2!

EzraBrooks:
**Public API Changes**

None.

**Description**

- Adds support for testing against ROS 2
- Adds humble, jazzy, and kilted to the test matrix in CI
- Disables ROS-1-only tests when running against distros other than `noetic`
- Rewrote some tests that were way too timing-sensitive. Changed the way they asserted on their output, but left the assertions the same.

Closes #746

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,6 @@ jobs:
           - humble
           - jazzy
           - kilted
-          - rolling
         node_version: [20, 22, 24]
     env:
       ROS_DISTRO: ${{ matrix.ros_distro }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ros_distro: [noetic]
+        ros_distro: [noetic, humble, jazzy]
         node_version: [20, 22, 24]
     env:
       ROS_DISTRO: ${{ matrix.ros_distro }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ros_distro: [noetic, humble, jazzy]
+        ros_distro:
+          - noetic
+          - humble
+          - jazzy
+          - kilted
+          - rolling
         node_version: [20, 22, 24]
     env:
       ROS_DISTRO: ${{ matrix.ros_distro }}
--- a/test/examples/check-topics.example.ts
+++ b/test/examples/check-topics.example.ts
@@ -1,18 +1,7 @@
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi } from "vitest";
 import * as ROSLIB from "../../src/RosLib.js";
 
-const expectedTopics = [
-  /*
-   * '/turtle1/cmd_vel', '/turtle1/color_sensor', '/turtle1/pose',
-   * '/turtle2/cmd_vel', '/turtle2/color_sensor', '/turtle2/pose',
-   */
-  "/tf2_web_republisher/status",
-  "/tf2_web_republisher/feedback",
-  // '/tf2_web_republisher/goal', '/tf2_web_republisher/result',
-  "/fibonacci/feedback",
-  "/fibonacci/status",
-  "/fibonacci/result",
-];
+const expectedTopics = ["/listener"];
 
 describe("Example topics are live", function () {
   const ros = new ROSLIB.Ros({
@@ -55,18 +44,12 @@ describe("Example topics are live", function () {
       });
     }));
 
-  it(
-    "unadvertise will end the topic (if it's the last around)",
-    () =>
-      new Promise((done) => {
-        console.log("Unadvertisement test. Wait for 15 seconds..");
-        setTimeout(function () {
-          ros.getTopics(function (result) {
-            expect(result.topics).not.to.contain("/some_test_topic");
-            done(result);
-          });
-        }, 15000);
-      }),
-    20000,
-  );
+  it("unadvertise will end the topic (if it's the last around)", async () => {
+    console.log("Unadvertisement test. Wait for 15 seconds..");
+    vi.waitFor(function () {
+      ros.getTopics(function (result) {
+        expect(result.topics).not.to.contain("/some_test_topic");
+      });
+    }, 15000);
+  });
 });
--- a/test/examples/check-topics.example.ts
+++ b/test/examples/check-topics.example.ts
@@ -8,48 +8,50 @@ describe("Example topics are live", function () {
     url: "ws://localhost:9090",
   });
 
-  it("getTopics", () =>
-    new Promise((done) => {
-      ros.getTopics(function (result) {
-        expectedTopics.forEach(function (topic) {
-          expect(result.topics).to.contain(
-            topic,
-            "Couldn't find topic: " + topic,
-          );
-        });
-        done(result);
-      });
-    }));
+  it("getTopics", async () => {
+    const callback = vi.fn();
+    ros.getTopics(callback);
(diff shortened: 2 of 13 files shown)


test/examples/check-topics.example.ts
```diff
@@ -1,18 +1,7 @@
...
-   * '/turtle1/cmd_vel', '/turtle1/color_sensor', '/turtle1/pose',
-   * '/turtle2/cmd_vel', '/turtle2/color_sensor', '/turtle2/pose',
-   */
-  "/tf2_web_republisher/status",
-  "/tf2_web_republisher/feedback",
-  // '/tf2_web_republisher/goal', '/tf2_web_republisher/result',
-  "/fibonacci/feedback",
-  "/fibonacci/status",
-  "/fibonacci/result",
-];
```
EzraBrooks, line 15: This test data is ROS-1-only and wasn't even sensical.. it didn't check for _the one thing unique to our test suites_, the `/listener` topic 🤦🏻‍♂️

EzraBrooks:
Test failure that only happens on first run :eyes: :eyes: :eyes:

EzraBrooks:
there seems to be a determinism issue in the pubsub example..

EzraBrooks:
I borked something... #1025

EzraBrooks:
https://github.com/RobotWebTools/roslibjs/actions/runs/19199950081/job/54886332135?pr=1024#step:6:82

Caught the malformed JSON bug in the wild in one of the Humble jobs @bjsowa @sea-bass

.github/workflows/main.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ros_distro: [noetic]
```
bjsowa, line 20: Why not test klited and rolling as well?

EzraBrooks: Sure, we can expand this matrix, but first let's get Humble and Jazzy working. Right now, they're failing differently, which makes me suspect rosbridge

EzraBrooks: ok, after rewriting the test that was failing on both ROS 2 distros (because it was relying on deterministic Topic behavior that isn't present in ROS 2 without EventsExecutor), now Humble is the only one failing. I actually think maybe some code got contributed to roslibjs that is incompatible with Humble, so I'll investigate that.

bjsowa:
@EzraBrooks It seems some tests related to Parameters are failing. Just FYI, I reimplemented the `params` module in `rosapi` at some point (RobotWebTools/rosbridge_suite#1001) but did not backport it to humble, so they may behave slightly different.

EzraBrooks:
~~Now we're into race-condition-land, because some of these Humble tests pass locally on my machine. I'll try cranking up their timeouts, since the default runners on GitHub Actions are tiny dual-core things.~~

edit: turns out the `getParams` call was hanging the bridge, which was failing random other tests.

test/examples/params.example.ts
```diff
@@ -1,63 +1,67 @@
...
+  });
+
+  it("Param.get", async () => {
+    const callback = vi.fn();
+    param.get(callback);
+    await vi.waitFor(() => expect(callback).toHaveBeenCalledWith(false));
+  });
+
+  // Known issue with getting params being able to hang in Humble because it had no timeout.
+  // This doesn't even work with `ros2 param list` at the CLI in our test environment :(
```
EzraBrooks, line 46: fyi

test/examples/setup_examples_ros2.launch.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<launch>
+  <!-- Include rosbridge websocket launch file for ROS 2 -->
+  <include file="$(find-pkg-share rosbridge_server)/launch/rosbridge_websocket_launch.xml">
+    <arg name="call_services_in_new_thread" value="true"/>
```
EzraBrooks, line 6: turns out our tests accidentally depend on this behavior which was not default in humble

EzraBrooks:
traded passing tests on humble for failing tests on noetic, excellent...

EzraBrooks:
With how much duplicated effort is going on here, I think roslibjs 2.x should be the final major release to support ROS 1.

I think there are a lot of great improvements in 2.0 that we should provide the ROS 1 users, but we should move on pretty quickly to a 3.0 that has no ROS 1 support.

EzraBrooks:
I enabled testing on rolling and then immediately disabled it because something outside my control was broken. Sorry, not gonna happen - at least not right now.

sea-bass:
> I enabled testing on rolling and then immediately disabled it because something outside my control was broken. Sorry, not gonna happen - at least not right now.

It just seems like whatever ros version is being used in the backend for the rolling test doesn't contain this change

https://github.com/ros2/rosidl/pull/887

sea-bass:
It honestly may get easily fixed by doing an apt-get upgrade here:

https://github.com/RobotWebTools/roslibjs/blob/76fc87e61e6bb7a9130710c3dc09b6ff95c93811/Dockerfile#L9

Although that has implications for caching, of course. Ideally the docker image is just up to date lol

https://hub.docker.com/layers/library/ros/rolling-ros-base/images/sha256-c61fdc63cea06cdd8f60b18946b68f880d874335e1d20cc0d1314dda5fbc8753

EzraBrooks:
I'm still not a fan of having my code be tested against the unstable branch of someone else's code, regardless.

sea-bass:
> I'm still not a fan of having my code be tested against the unstable branch of someone else's code, regardless.

The usual model (at least that makes sense to me) is to have rolling in CI so you get the early feedback, but not having it be a required check to pass

test/examples/params.example.ts
```diff
@@ -1,63 +1,67 @@
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi } from "vitest";
 import * as ROSLIB from "../../src/RosLib.js";
 
+const PARAM_NAME =
+  process.env.ROS_DISTRO === "noetic"
+    ? "/test/foo"
+    : // Is it crazy to muck around with use_sim_time here? I feel like it shouldn't matter..
```
sea-bass, line 7: Fine IMO

EzraBrooks:
And therein lies the problem where people in the ROS community become inoculated to the sight of failing CI jobs.

imo, next steps for the tests in this repo are to eliminate all the dependencies other than `rosbridge_websocket` and `tf2_web_republisher`, and then maybe at that point we could test against rolling since they're "our code".

test/examples/setup_examples_ros2.launch.xml
sea-bass: Maybe leave a comment related to this so that when humble goes eol someone can see this and delete it

I'd also enable the action equivalent param `send_action_goals_in_new_thread` (going off memory) for the same reason

sea-bass (approved):
woof, you are right the ros 1/2 branching logic is something we should seek to get rid of. Nice job figuring all that out though.

sea-bass:
> imo, next steps for the tests in this repo are to eliminate all the dependencies other than `rosbridge_websocket` and `tf2_web_republisher`, and then maybe at that point we could test against rolling since they're "our code".

These packages won't be able to easily stop depending on "core" libraries like `rclpy` or `rosidl` (which is what is failing atm), so this might not be super feasible.

But you're right, I don't think it's unreasonable to simply not test rolling. It means less risk of "red is okay" but also means by the time a new stable distro is cut the number of surprises has potentially amassed.

EzraBrooks:
I think it's easier to squash all the bugs at once than it is to be stressed out from random CI fires for a year ;)

PR #1026, opened 2025-11-09
Make all failureCallbacks default to console.error

EzraBrooks:
**Public API Changes**

None, but you'll get logs when things fail now!

**Description**

Inspired by the PR in which I'm wrestling with unit tests..

Also cleaned up a few weirdo callback logic cases with the help of a typescript-eslint rule banning unnecessary conditions.

### DIFF
--- a/src/actionlib/ActionClient.ts
+++ b/src/actionlib/ActionClient.ts
@@ -26,7 +26,7 @@ export default class ActionClient<
 > extends EventEmitter<{
   timeout: void;
 }> {
-  goals: Record<string, Goal<TGoal>> = {};
+  goals: Partial<Record<string, Goal<TGoal>>> = {};
   /** flag to check if a status has been received */
   receivedStatus = false;
   ros: Ros;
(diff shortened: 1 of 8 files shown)


src/actionlib/ActionClient.ts
```diff
@@ -26,7 +26,7 @@ export default class ActionClient<
 > extends EventEmitter<{
   timeout: void;
 }> {
-  goals: Record<string, Goal<TGoal>> = {};
+  goals: Partial<Record<string, Goal<TGoal>>> = {};
```
EzraBrooks, line 29: stopgap fix to make this not violate the no-unnecessary-conditions rule until we get around to enabling `noUncheckedIndexedAccess` in our tsconfig which is still another can of worms

src/tf/BaseTFClient.ts
```diff
@@ -7,9 +7,8 @@ import { std_msgs } from "../types/std_msgs.js";
  * Base class for TF Clients that provides common functionality.
  */
 export default class BaseTFClient {
-  frameInfos: Record<
-    string,
-    { transform?: Transform; cbs: ((tf: Transform) => void)[] }
+  frameInfos: Partial<
```
EzraBrooks, line 10: same thing about `noUncheckedIndexedAccess` here

sea-bass (approved):

PR #1027, opened 2025-11-09
Make tests fail when they log errors

EzraBrooks:
**Public API Changes**

None

**Description**

To ensure that callback failures don't sneak through in testing.

Also to ensure we aren't crowding users' consoles during normal operation.

### DIFF
--- a/test/setup/per-suite-setup.ts
+++ b/test/setup/per-suite-setup.ts
@@ -1,0 +1,5 @@
+import failOnConsole from "vitest-fail-on-console";
+
+failOnConsole({
+  shouldFailOnWarn: false,
+});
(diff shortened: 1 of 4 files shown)


test/setup/per-suite-setup.ts
```diff
@@ -0,0 +1,5 @@
+import failOnConsole from "vitest-fail-on-console";
+
+failOnConsole({
+  shouldFailOnWarn: false,
```
EzraBrooks, line 4: This is to allow the pre-existing `console.warn` invocation in the CBOR implementation.

sea-bass (approved):
sweet

PR #1028, opened 2025-11-09
Fix types of shorthand constructors

EzraBrooks:
**Public API Changes**

None, just types the existing implicit signature.

**Description**

Does what it says on the tin.


sea-bass (approved):
👍🏻

PR #1030, opened 2025-11-10
Prepare for monorepo

EzraBrooks:
**Public API Changes**

None

**Description**

Moving everything `roslib`-related into `packages/roslib` to prepare for providing more packages, like `ros2djs`, `ros3djs`, and/or a first-party React binding to replace the removed React examples.


EzraBrooks:
In a follow-up PR I'll create an [npm workspace](https://docs.npmjs.com/cli/v7/using-npm/workspaces?v=true) at the root of the repository to make building and testing all packages in the workspace easy (`npm run test --workspace`), but I wanted to make sure that the existing `package.json` and `package-lock.json` get recorded as a rename instead of showing a diff where they look like they were deleted and replaced with the content of the workspace file.

EzraBrooks:
neat. gonna leave this as a draft until we have consensus on whether this is the way forward.

And also probably until we release 2.0.0 because I don't want to make it hard for people to use the package via a git URL, which is the recommended solution until we ship 2.0.0, since NPM doesn't support installing from a subdirectory in a git URL.

EzraBrooks:
I think I'd like to merge this past failing CI (once we're ready) so that in an immediately following PR I can create a brand new package.json at the root of the repository without messing up the diff of the existing package.json.

EzraBrooks:
@bjsowa CI is failing on purpose here because I'm playing `git blame` games. it passes in #1093 which will be merged immediately afterward

bjsowa (approved):

PR #1033, opened 2025-11-10
Upgrade strictness of typescript eslint

EzraBrooks:
Bumped typescript-eslint all the way up to "strict", which found lots of weirdness. Some I introduced during the TypeScript port, some that already existed and needed to be annotated correctly.

Fixed it all.

Not sure if we'll keep it at strict in the long term (because that might make it hard for non-TypeScript-experts to contribute), we could bump it down to "recommended type checked" later, but I think having it be strict for awhile might be a good thing to try.

### DIFF
--- a/src/actionlib/Goal.ts
+++ b/src/actionlib/Goal.ts
@@ -18,7 +18,7 @@ export default class Goal<
   TFeedback = unknown,
   TResult = unknown,
 > extends EventEmitter<{
-  timeout: void;
+  timeout: undefined;
   status: [actionlib_msgs.GoalStatus];
   feedback: [TFeedback];
   result: [TResult];
@@ -28,7 +28,7 @@ export default class Goal<
   result?: TResult = undefined;
   feedback?: TFeedback = undefined;
   // Create a random ID
-  goalID = "goal_" + Math.random() + "_" + new Date().getTime();
+  goalID = `goal_${Math.random().toString()}_${new Date().getTime().toString()}`;
   actionClient: ActionClient<TGoal, TFeedback, TResult>;
   goalMessage: { goal: TGoal; goal_id: actionlib_msgs.GoalID };
   /**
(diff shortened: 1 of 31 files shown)


src/actionlib/Goal.ts
```diff
@@ -18,17 +18,17 @@ export default class Goal<
...
   result: [TResult];
 }> {
   isFinished = false;
-  status = undefined;
+  status?: actionlib_msgs.GoalStatus = undefined;
   result?: TResult = undefined;
   feedback?: TFeedback = undefined;
   // Create a random ID
-  goalID = "goal_" + Math.random() + "_" + new Date().getTime();
+  goalID = `goal_${Math.random().toString()}_${new Date().getTime().toString()}`;
```
bjsowa, line 31: Not related to the scope of this PR but I find it weird to use a floating point value for a random identifier. Maybe something like UUID would be more appropriate.

EzraBrooks: Totally agreed, could you open an issue for that?

src/core/Ros.ts
```diff
@@ -689,9 +702,9 @@ export default class Ros extends EventEmitter<
           }
         } else {
           // lookup the name
-          let sub = false;
+          let sub: boolean | rosapi.TypeDef = false;
```
bjsowa, line 705: Wouldn't it be more appropriate to use something like this for an optional type?
```suggestion
          let sub: rosapi.TypeDef | undefined;
```

src/actionlib/Goal.ts
bjsowa: #1034

bjsowa (approved):
I'm generally a fan of strict static type checking. That's why I adopted it in Rosbridge. Might be a bit harder to contribute but helps preventing many bugs

src/core/Ros.ts
EzraBrooks: Agreed; this weird "or false" pattern in the codebase should be refactored out. I should open an issue about finding and replacing all these types.

EzraBrooks: #1035

PR #1036, opened 2025-11-10
Remove boolean unions

### LINKED ISSUE #1035: Factor out Boolean union types in favor of optional

EzraBrooks:
Closes #1035 

turns out there were only two places left @bjsowa 😆

### DIFF
--- a/src/core/Ros.ts
+++ b/src/core/Ros.ts
@@ -702,7 +702,7 @@ export default class Ros extends EventEmitter<
           }
         } else {
           // lookup the name
-          let sub: boolean | rosapi.TypeDef = false;
+          let sub: rosapi.TypeDef | undefined = undefined;
           for (const hint of hints) {
             if (hint.type === fieldType) {
               sub = hint;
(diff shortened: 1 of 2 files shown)


src/core/Ros.ts
```diff
@@ -702,7 +702,7 @@ export default class Ros extends EventEmitter<
           }
         } else {
           // lookup the name
-          let sub: boolean | rosapi.TypeDef = false;
+          let sub: rosapi.TypeDef | undefined = undefined;
```
bjsowa, line 705: Is rosapi.TypeDef guaranteed to evaluate to true?

EzraBrooks: Yep. Even an empty object is "truthy" in JavaScript, and I don't think it's possible to get `undefined`, `false`, or `null` over the wire here. I guess it's possible if you're using a bridge that isn't implemented properly, but I don't think we need to account for that.

bjsowa (approved):

PR #1037, opened 2025-11-10
Replace ID counter with UUIDs

### LINKED ISSUE #1034: Don't use Math.random() as unique identifier
I find it weird to use a floating point value for a random identifier. Maybe something like UUID would be more appropriate.

_Originally posted by @bjsowa in https://github.com/RobotWebTools/roslibjs/pull/1033#discussion_r2510776363_

EzraBrooks:
Closes #1034

### DIFF
--- a/src/core/Action.ts
+++ b/src/core/Action.ts
@@ -12,6 +12,7 @@ import {
   RosbridgeSendActionGoalMessage,
 } from "../types/protocol.ts";
 import Ros from "./Ros.js";
+import { v4 as uuidv4 } from "uuid";
 
 /**
  * A ROS 2 action client.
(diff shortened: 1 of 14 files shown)


bjsowa (approved):

EzraBrooks:
why did this break noetic 😆

EzraBrooks:
oh! clobbered goal IDs from multiple ROS instances in the tests!

We should probably just switch everything to UUIDs.

src/core/Action.ts
```diff
@@ -67,7 +68,7 @@ export default class Action<
       return;
     }
 
-    const actionGoalId = `send_action_goal:${this.name}:${(++this.ros.idCounter).toString()}`;
+    const actionGoalId = `send_action_goal:${this.name}:${uuidv4()}`;
```
bjsowa, line 71: Can we replace other places to also use template literals instead of string concatenation, for consistency?

EzraBrooks: done via an eslint rule

PR #1039, opened 2025-11-10
Add bson_only_mode integration tests

### LINKED ISSUE #1038: Add integration tests for bson-only mode

EzraBrooks:
Closes #1038

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ defaults:
 
 jobs:
   ci:
-    name: ${{ matrix.ros_distro }} (node ${{ matrix.node_version }})
+    name: ${{ matrix.ros_distro }} (node ${{ matrix.node_version }}) (bson ${{ matrix.bson_only_mode }})
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
@@ -23,8 +23,12 @@ jobs:
           - jazzy
           - kilted
         node_version: [20, 22, 24]
+        bson_only_mode:
+          - true
+          - false
     env:
       ROS_DISTRO: ${{ matrix.ros_distro }}
+      BSON_ONLY_MODE: ${{ matrix.bson_only_mode }}
     steps:
       - uses: actions/checkout@v5
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 ARG ROS_DISTRO=noetic
 FROM ros:${ROS_DISTRO}-ros-base
+ARG BSON_ONLY_MODE=false
+
+ENV BSON_ONLY_MODE=${BSON_ONLY_MODE}
 
 # Copy package.xml and install ROS dependencies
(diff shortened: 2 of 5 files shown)


EzraBrooks:
welp, turns out BSON doesn't work.

bjsowa:
> welp, turns out BSON doesn't work.

I kinda expected it since there are no tests for BSON in rosbridge as well.

EzraBrooks:
I have a creeping fear that neither BSON nor CBOR nor PNG work.

I think I understand the problem here, though - the `Ros` class doesn't know that the server is in BSON-only mode. Is there an API exposed that roslibjs could query to ask that?

EzraBrooks:
I might argue that we take a loss on the current support matrix of encodings and compressions and we design a breaking change to the API that consolidates everything to one Blessed standard compression.

Right now, we have a matrix of:

|    |JSON|BSON|
|----|----|----|
|Raw |✔️  |✔️  
|CBOR|✔️  |✔️
|CBOR-RAW| ✔️| ✔️ 
|PNG |✔️  |✔️

Which effectively means each server implementation and each client implementation has to maintain **eight** versions of the protocol. We have more protocols than maintainers! I think we could kill BSON, PNG, and Raw entirely and use CBOR over the wire by default.

bjsowa:
There's also [CBOR-RAW encoding](https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/ROSBRIDGE_PROTOCOL.md#314-cbor-raw-encoding--cbor-raw-)

EzraBrooks:
So we have eight protocols 😆 

We definitely need to de-scope. maybe that's roslibjs 3.x and rosbridge 4.x?

bjsowa:
> So we have eight protocols 😆
> 
> We definitely need to de-scope. maybe that's roslibjs 3.x and rosbridge 4.x?

I'm hesitant to cut out features. There are multiple other rosbridge client implementations and who knows what they might use. If it happens that there is no real advantage of using BSON instead of CBOR, I might be ok with cutting it in Rolling and upcoming L-turtle distribution.

EzraBrooks:
Out of the various options it does seem like BSON is the easiest to cut, since BSON has largely fallen out of fashion due to the fact that in many cases it's actually less efficient than JSON. And it's a top-level encoder for us, which makes it combinatorial - removing it would take us from eight dialects to four.

EzraBrooks:
I would argue that we should go way further than just cutting one option for the next distro, though - we could go as far as switching everything to LZMA-compressed JSON by default.

However, I understand your hesitancy, since the ROS distribution model makes it hard for your consumers to avoid downstream effects since they can't easily install old versions.

I will say that I think that if I manage to shore up roslibjs, ros2djs, and ros3djs into good states over the coming months, I may consider resurrecting `ros2-web-bridge` to be able to have a bridge implementation that can be installed via NPM at a specific matched version.

PR #1042, opened 2025-11-10
Add transport factory option to customize the websocket used

douglascayers:
**Public API Changes**

When creating a `new Ros(options)` instance, expanded the transport library option to accept a factory function.

**Description**

We recently updated from Node 20 to Node 22 and noticed that the Ros websocket connection behavior was different. If the server was unstable and the socket could not connect immediately, when we were on Node 20 then the Ros websocket would repeatedly try until connected. But on Node 22 there was a single attempt and then the code seemed to hang.

This came down to [Node 22 now ships](https://nodejs.org/en/learn/getting-started/websocket) with a [native WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) impl which behaves differently than the [ws package](https://www.npmjs.com/package/ws).

We would like to continue using the `ws` package rather than the native WebSocket class.

**Example Usage**
```ts
import * as roslib from 'roslib';

// same original behavior
// will use native WebSocket if available else ws package
new roslib.Ros();

// same original behavior
// will use native WebSocket if available else ws package
new roslib.Ros({
  transportLibrary: "websocket",
});

// new opt-in behavior to create your own websocket instance
new roslib.Ros({
  transportLibrary: async (url: string): Promise<roslib.RequiredSocketInterface> => {
    // ...
  },
});

// example using ws package with custom options
import WebSocket from 'ws';

new roslib.Ros({
  transportLibrary: async (url: string): Promise<roslib.RequiredSocketInterface> => {
    return new WebSocket(url, {
      handshakeTimeout: 30_000,
    });
  },
});
```

### DIFF
--- a/src/core/Ros.ts
+++ b/src/core/Ros.ts
@@ -35,7 +35,7 @@ export type TransportLibrary =
   | RTCPeerConnection
   | TransportFactory;
 
-export interface RTCTransportOptions extends RTCDataChannelInit {
+export interface TransportOptions extends RTCDataChannelInit {
   /**
    * Hook to handle outgoing socket message and encode it.
    */
@@ -81,7 +81,7 @@ export default class Ros extends EventEmitter<
   socket: SocketAdapter | null = null;
   isConnected = false;
   transportLibrary: TransportLibrary;
-  transportOptions: RTCTransportOptions;
+  transportOptions: TransportOptions;
   /**
    * @param [options]
    * @param [options.url] - The WebSocket URL for rosbridge. Can be specified later with `connect`.
@@ -95,7 +95,7 @@ export default class Ros extends EventEmitter<
   }: {
     url?: string;
     transportLibrary?: TransportLibrary;
-    transportOptions?: RTCTransportOptions;
+    transportOptions?: TransportOptions;
   } = {}) {
     super();
 
--- a/src/core/Ros.ts
+++ b/src/core/Ros.ts
@@ -3,7 +3,8 @@
  * @author Brandon Alexander - baalexander@gmail.com
  */
 
-import socketAdapter from "./SocketAdapter.js";
+import type { RequiredSocketInterface } from "./SocketAdapter.js";
+import SocketAdapter from "./SocketAdapter.js";
 import type { RosbridgeMessage } from "../types/protocol.js";
 import {
   isRosbridgeActionFeedbackMessage,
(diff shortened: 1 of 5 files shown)


EzraBrooks:
Thanks for the contribution!

Did you investigate if it's possible to pass arguments to the native websocket that make it emulate the ws behavior you like? It would be great to have consistent behavior between browser and Node.

douglascayers:
> Thanks for the contribution!
> 
> Did you investigate if it's possible to pass arguments to the native websocket that make it emulate the ws behavior you like? It would be great to have consistent behavior between browser and Node.

I looked at the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket) but didn't readily see such an option to behave like the `ws` package.

Also happy to rework this approach to something else you prefer.

My preferred goals are:
1. Tell roslib which websocket impl to use
2. If we can do (1), maybe pass down specific options for that impl

At this time, I really only need (1). I thought about proposing a factory function whose default behavior would be `this.createTransport(url)` but then saw that the factory needs logic around when it's appropriate to create a new transport (websocket) instance or not and thought ok, maybe that's leaking too many details.

EzraBrooks:
I actually only just extracted the `createTransport` function this past week - it's definitely not an API I'm attached to. It was to get rid of the previous hard-to-decipher behavior of effectively spreading the socketadapter into a constructed websocket.

If we could find a way to simplify createTransport to take a factory and avoid the kind of scary union type that currently exists to accept either a string or an RTC connection, that would probably be a huge clarity improvement. And as long as the default implementation is the same as the current default I think I'd be happy breaking that API for 2.0 (which I'm shipping any day now)

douglascayers:
> If we could find a way to simplify createTransport to take a factory and avoid the kind of scary union type that currently exists to accept either a string or an RTC connection, that would probably be a huge clarity improvement.

On it, boss!

Will refactor this PR to support a factory method instead, which may have added benefit of letting users choose other websocket libraries if they wish

EzraBrooks:
Sweet! The type I just added to SocketAdapter last night(?) that describes the minimal required interface for roslibjs to communicate might be useful here.

douglascayers:
@EzraBrooks I've refactored the PR to instead introduce a transport factory function. Also includes updated tests for SocketAdapter and Ros classes.

EzraBrooks (commented):
Great work! The fact that you didn't have to modify any of the existing tests for the classes other than SocketAdapter is evidence enough to me that you've preserved the most common user workflow :)

I wonder if we should go a step further and remove the RTC logic now that a user can pass their own RTC connection factory, since I've never heard of anyone using it and it's not integration-tested. Then `TransportLibrary` could just be `TransportFactory | undefined`. I don't think it's likely that many people were specifying `"websocket"` as its value since that was the default.

EzraBrooks:
It seems like [this](https://github.com/RobotWebTools/webrtc_ros) may have been the intended use of WebRTC, which hasn't received a substantial commit in.. 4 years 😬

bjsowa:
> It seems like [this](https://github.com/RobotWebTools/webrtc_ros) may have been the intended use of WebRTC, which hasn't received a substantial commit in.. 4 years 😬

I think this project is unrelated to rosbridge, it was just an alternative to `web_video_server`.

EzraBrooks:
The only evidence we have of usage is #223, which has _no details whatsoever_, so it's entirely possible this never actually worked with any publicly-available implementation of RWT.

douglascayers:
> I wonder if we should go a step further and remove the RTC logic now that a user can pass their own RTC connection factory, since I've never heard of anyone using it and it's not integration-tested. Then TransportLibrary could just be TransportFactory | undefined. I don't think it's likely that many people were specifying "websocket" as its value since that was the default.

I like the idea of decoupling the Ros class from knowing anything about the socket implementations.

Happy to continue iterating on this per those suggestions.

In the chance someone _was_ using RTC, we could provide a couple factory functions for people, or just provide documentation on how to it. Either way, the Ros core would not be bound to the implementations but rather just continue to operate on an "socket interface" of minimally required methods/props that SocketAdapter requires.

EzraBrooks:
Given someone asked about this in the past (#798), I think it probably makes sense to just have an abstract socket interface and provide reference implementations for ws and WebSocket, defaulting to the latter. Someone should be able to infer from those reference implementations how to make one for RTC or whatever else!

douglascayers:
> just have an abstract socket interface and provide reference implementations for ws and WebSocket, defaulting to the latter

Agree, will make it happ'n cap'n 🫡

douglascayers:
> Given someone asked about this in the past (#798), I think it probably makes sense to just have an abstract socket interface and provide reference implementations for ws and WebSocket, defaulting to the latter. Someone should be able to infer from those reference implementations how to make one for RTC or whatever else!

Because that change was larger in scope than this PR, I've made that proposal in https://github.com/RobotWebTools/roslibjs/pull/1060

If you like that one, it can supersede this one and we close this one.

Thanks!

EzraBrooks:
agreed #1060 is a great path forward!

PR #1044, opened 2025-11-10
Enable all the ts checks we're now passing

EzraBrooks:
Moves a bunch of our TODO typescript checks into the enabled section because we're now passing them due to refactors done on the way during the migration.


bjsowa (commented):
Why was the order of the options changed?

tsconfig.build.json
EzraBrooks, line 28: @drewhoener had broken it up into sections of "rules that we have enabled" and "rules we don't" because in some cases the strict mode is `false` and in some cases it's `true`. I was just following that pattern.

Once everything is strict all these options can be removed in favor of setting `strict: true`.

bjsowa: Oh, yeah, sorry. Didn't notice it was split like that. That makes sense

bjsowa (approved):

PR #1047, opened 2025-11-10
Enable `strictFunctionTypes`

EzraBrooks:


src/core/SocketAdapter.ts
```diff
@@ -65,7 +65,7 @@ export default class SocketAdapter {
     }: {
       onOpen: (event: Event) => void;
       onClose: (event: Event) => void;
-      onError: (event: ErrorEvent) => void;
+      onError: (event: ErrorEvent | RTCErrorEvent) => void;
```
bjsowa, line 68: Kinda weird that `RTCErrorEvent` does not inherit from `ErrorEvent`

bjsowa (approved):

PR #1056, opened 2025-11-11
Finally commit to breaking the module API for 2.0

EzraBrooks:
Friendship ended with barrel files and a top-level default export. Traceable named imports are my best friend now.

**Public API Changes**

- All users will have to `import * as ROSLIB from "roslib"` if they were previously using `import ROSLIB from "roslib"`.
- ROSLIB assigning itself to your global window state at startup if you're running it from a CDN is gone. Since our plan for migrating CDN users to 2.0 was already to point them in the direction of UNPKG, [we can point them in the direction of UNPKG's documentation of import maps](https://unpkg.com/#nobuild-apps) to address this API difference.

**Description**

I removed all index files in the name of improving the tree-shakeability of our codebase and our consumers' codebases - if you're importing everything, you can't shake anything out.


EzraBrooks:
Why can't you put a PR back to draft status from the mobile app..

EzraBrooks:
I forgot there was a test checking that exactly this behavior is present 😆

bjsowa (approved):

PR #1057, opened 2025-11-12
Stop bundling dependencies, UMD file

EzraBrooks:
Allow users' package managers to satisfy our dependencies, instead of bundling them all in our compiled output.

(this is like dynamic linking as opposed to static linking for JS)

As an aside, the question now becomes "why use a bundler like Vite at all now when we're not really 'bundling' anything?". Why indeed? I ported us to Vite to replace a bunch of old tooling, but maybe we don't even need Vite in the long term since the JavaScript module-importing ecosystem is far more mature than it once was.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,7 @@
         "vite": "^7.0.0",
         "vite-plugin-checker": "^0.11.0",
         "vite-plugin-dts": "^4.0.2",
+        "vite-plugin-externalize-deps": "^0.10.0",
         "vitest": "^4.0.3",
         "vitest-fail-on-console": "^0.10.1"
       },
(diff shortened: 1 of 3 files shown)


EzraBrooks:
We've already been forcing all our package-manager-using users to download all our dependencies, and then we weren't actually using them. This PR takes our compiled library size down from over 300kB to less than 70kB. As I mentioned in #1056, CDN users (however many of them will actually migrate to 2.0, given that CDN-based JS usage is pretty old-fashioned) will just need to set up an import map - which is not unlike how we used to tell users to grab EventEmitter from a CDN in addition to grabbing roslibjs.

bjsowa:
@Krzemien97 Does this affect us in any way?

EzraBrooks:
As long as you guys are installing via a package manager like npm, pnpm, yarn, etc. and are using an ESM-compatible build tool (basically all of them at this point), it shouldn't.

bjsowa (approved):

PR #1058, opened 2025-11-12
Use typedoc warnings to find and fix partially-hidden APIs

EzraBrooks:
Used typedoc's warnings about unexported types being referred to in the public API to find a bunch of stuff that should be either exported or made private. Some stuff ended up being deleted because when it was made private it had no references!

## Public API changes

A bunch of stuff that I _believe_ to be implementation details got moved from public to private. Please double-check me, but clearly none of the stuff was used in the examples or integration tests since those are still building.

I also changed a couple signatures in the math library to have their members be `number | undefined` instead of `number | null` because the existing API requiring users to specify `null` was lame and probably wasn't actually even being used that way in practice before strict typing.


src/actionlib/ActionClient.ts
```diff
@@ -36,17 +36,17 @@ export default class ActionClient<
...
     feedback: TFeedback;
   }>;
-  statusListener: Topic<actionlib_msgs.GoalStatusArray>;
-  resultListener: Topic<{
+  #statusListener: Topic<actionlib_msgs.GoalStatusArray>;
+  #resultListener: Topic<{
     status: actionlib_msgs.GoalStatus;
     result: TResult;
   }>;
   goalTopic: Topic<{ goal: TGoal; goal_id: actionlib_msgs.GoalID }>;
```
EzraBrooks: This had to stay public despite being an implementation detail because of some wonky stuff where `Goal` refers to `this.actionClient.goalTopic` :(

EzraBrooks: actually, we won't have a better time to break this API than 2.0, so I'm gonna add another function to hide this so it can be private.

EzraBrooks:
Superseded by #1073. I figure changing the API of actionlib at this point doesn't actually make sense since extracting it from the core package is one of the first things I'll do after cutting 2.0.

PR #1059, opened 2025-11-12
Import files with .ts extensions

EzraBrooks:
**Public API Changes**

None

**Description**

We have at least one passionate user (because they contributed!) using Bun, so there's probably others. Bun and Deno can run .ts files directly - but if we confuse them with import paths that don't resolve until after a transpile step, they can break.

This PR ensures all files are imported as .ts files for compatibility with TypeScript runtimes and then rewritten at build time for compatibility with traditional JavaScript runtimes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "dockerode": "^4.0.9",
         "eslint": "^9.32.0",
         "eslint-config-prettier": "^10.1.8",
+        "eslint-plugin-import": "^2.32.0",
         "eslint-plugin-jsdoc": "^61.1.12",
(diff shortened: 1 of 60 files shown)


package-lock.json
EzraBrooks, line 1: jeez, 2000 lines added to package-lock. `eslint-plugin-import` has so many dependencies. good thing it's a dev dependency and won't ship to users.

EzraBrooks: I do want this package for other reasons too, though, so it's a price I'm willing to pay.

bjsowa (approved):

PR #1060, opened 2025-11-13
Replace SocketAdapter with Transports

douglascayers:
### Background

A variation of https://github.com/RobotWebTools/roslibjs/pull/1042 based on the comment thread there about extending the factory function concept further to abstract the Ros class from socket implementations.

### Changes

- Replaces `SocketAdapter` with `Transport`, which provide an abstraction over socket implementations (native WebSocket, ws npm module, socket.io, RTC, etc).
- Removed RTC web socket implementation. Users who need it can provide their own `TransportFactory` modeled after the `WebSocketTransportFactory`.
- Errors when handling socket messages to be parsed into `RosbridgeMessage` are now emitted by the transport rather than console logged or swallowed. Gives calling code an easier programmatic option to be notified of these issues.
- Updated type definitions for a few utilities.
- Add and update tests.
- Changed event `connection` to `open` to be consistent with the underlying socket event type.
- Ros emits the `open`, `close`, and `error` events as pass-throughs of whatever value the transport emitted (e.g. string, Error, Event, etc)

### Example Usage

```ts
// -- Example 1 -- //

// Defaults to WebSocket transports
// Uses native WebSocket if available, else the "ws" node module
const ros = new Ros();

// -- Example 2 -- //

// Customize the WebSocket implementation
// This example specifies the handshake timeout
const ros = new Ros({
  transportFactory: async (url: string): ITransport => {
    const socket = new ws.WebSocket(url, {
      handshakeTimeout: 30_000,
    });
    socket.binaryType = "arraybuffer";
    return new WsWebSocketTransport(socket);
  },
});

// -- Example 3 -- //

// Bring your own transport with full control
// over encoding/decoding messages
const MyCustomTransport extends EventEmitter implements ITransport {
  constructor(url: string) {
    // connect to the url however you wish
    // when receive messages from the server,
    // broadcast them via `this.emit("message", ...)`
  }

  send(message: RosbridgeMessage): void {
    // roslib wants to send a message to the server
    // serialize and send the message however you wish
  }

  close(): void {
    // disconnect from the server
  }

  // lifecycle methods to mirror WebSocket spec
  isConnecting(): boolean { ... }
  isOpen(): boolean { ... }
  isClosing(): boolean { ... }
  isClosed(): boolean { ... }
}

const ros = new Ros({
  transportFactory: async (url: string): ITransport => {
    return new MyCustomTransport(url);
  },
});
```


test/setup/ros-backend.ts
```diff
@@ -18,7 +18,7 @@ const docker = new Docker();
  */
 function getRosDistro() {
   // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing -- this might be an empty string, not undefined.
-  return process.env["ROS_DISTRO"] || "noetic";
+  return (process.env["ROS_DISTRO"] ||= "noetic");
```
douglascayers, line 21: If the env var isn't set and the method defaults to `"noetic"`, the `||=` assignment also assigns that value to the env var now so that tests with `skipIf` logic will run if they are filtering for `"noetic"` distro (e.g. fibonacci)

src/util/decompressPng.ts
```diff
@@ -15,16 +15,33 @@ import pngparse from "pngparse";
...
+    callback(undefined, data);
+  };
+  const reject = (reason: unknown) => {
+    let message = "Cannot process PNG encoded message: ";
+    if (reason instanceof Error) {
+      message += reason.message;
+    } else {
+      message += String(reason);
+    }
+    callback(new Error(message));
```
EzraBrooks: I think you can just set { cause: error } in the constructor here instead of doing string manipulation.

src/util/shim/decompressPng.ts
```diff
@@ -13,8 +13,20 @@
...
+    callback(undefined, data);
+  };
+  const reject = (reason: unknown) => {
+    let message = "Cannot process PNG encoded message: ";
+    if (reason instanceof Error) {
+      message += reason.message;
+    } else {
+      message += String(reason);
+    }
+    callback(new Error(message));
```
EzraBrooks: Same comment about cause

test/setup/ros-backend.ts
EzraBrooks: Isn't this environment variable already set appropriately within the container by sourcing the ROS workspace on bringup? Assigning to process.env kind of skeeves me out.

src/util/decompressPng.ts
EzraBrooks: Actually, should we merge my PR fixing up the PNG logic and disable its test that is currently failing on ROS 2 (because it's a rosbridge issue not a roslibjs issue) and rebase this to simplify?

src/core/Transport.ts
```diff
@@ -0,0 +1,381 @@
...
+
+/**
+ * A transport factory that uses WebSockets to send and receive messages.
+ * Will use the native `WebSocket` class if available, otherwise falls back
+ * to the `ws` package.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/WebSocket
+ * @see https://github.com/websockets/ws
+ */
+export class WebSocketTransportFactory implements ITransportFactory {
```
EzraBrooks: Is there a reason this needs to be a class with a singular function member instead of just a free function?

eslint.config.ts
```diff
@@ -31,6 +31,19 @@ export default defineConfig(
       "@typescript-eslint/no-empty-function": 0,
       "prefer-template": 2,
       "@typescript-eslint/consistent-type-imports": 2,
+      // Disabled to allow some variables to be explicitly ignored
+      "@typescript-eslint/no-unused-vars": [
```
EzraBrooks: Noting to myself to look at why this was exempted during a second pass review - probably fine but I wanna double-check it's not papering over a weird API for users

EzraBrooks (commented):
Looks great, just some nits

EzraBrooks:
btw I'm planning on really highlighting this feature in the 2.0 release notes because it's cool and it will help anyone who was relying on the almost-certainly-unmaintained NodeTCP and RTC transports :)

test/setup/ros-backend.ts
douglascayers: I'll double check. I was trying to figure out why the fibonacci test was being skipped for me but maybe it was unrelated.

EzraBrooks: Oh, you know what, retracing my steps in my head you are totally right. This is a good stopgap fix, we should figure out a way that doesn't involve mutating process.env in the future though

src/util/decompressPng.ts
EzraBrooks: #980 is merging momentarily

test/setup/ros-backend.ts
douglascayers: It may be that the tests aren't running in the docker container but rather on the host machine where the ROS_DISTRO env var might not be set (unless the tester explicitly sets it)

I tried adding `ENV ROS_DISTRO=${ROS_DISTRO}` to the Dockerfile and rebuilt it without cache but the fibonacci test was still skipped.

eslint.config.ts
EzraBrooks: is this just for pngparse? because #980 eliminates pngparse :)

test/setup/ros-backend.ts
douglascayers: Another approach is have the tests' skipIf condition reference the `getRosDistro()` method from the docker backend if don't want the setup code to set the `process.env.ROS_DISTRO`. My personal vote is let the setup code set the ROS_DISTRO so the tests don't have to know about the docker setup code.

Image

EzraBrooks: > It may be that the tests aren't running in the docker container but rather on the host machine 

correct, I did this on purpose to not entangle the ROS env with the node env. I just happened to miss the detail of synchronizing this env variable :/

eslint.config.ts
douglascayers: I added it so that the `callOnConnection` method could specify that it accepts anything that extends from `RosbridgeMessage`. Otherwise, the linter was error'ing if I tried to pass in any object that had more properties than what strictly `RosbridgeMessage` defined.

Image

src/core/Transport.ts
douglascayers: I was thinking what if someone needed their factory to be smarter than just a function call? What if they wanted their factory to maintain some kind of state like a connection pool maybe or would limit how many simultaneous connections they wanted to allow or "insert some concept here". If people pass in an object then I think that makes it easier to extend for future capabilities we don't yet know we may want factories be responsible for, too.

And if someone just needs to pass in a function, then their factory can be as as simple as enclosing it in a small object.

```ts
const simpleFactory = {
  createTransport: (url: string): => { ... }
}

const complexFactory = {
  ... other state tracked here
  createTransport: (url: string): => { ... }
}

const factory2dot0 = {
  weWantAnotherMethodHere: () => { ... }
  createTransport: (url: string): => { ... }
}
```

At least, that was my thinking anyways :)

eslint.config.ts
EzraBrooks: Checking out this branch and removing the change to eslint.config doesn't fail any lint rules for me 🤔

src/core/Transport.ts
EzraBrooks: > if someone needed their factory to be smarter than just a function call

Couldn't they pass a reference to the function from the instantiated class in that case?

src/util/decompressPng.ts
douglascayers: Yes, I'm sure most or all of the changes I made for decompressPng will go away after I rebase, thanks

src/core/Transport.ts
douglascayers: I'll refactor the factory to be a function

eslint.config.ts
douglascayers: Interesting... I'm not getting the lint error for that method today either, very curious...

I'll remove it, thanks

src/core/Transport.ts
```diff
@@ -0,0 +1,379 @@
...
+  RosbridgePngMessage,
+  RosbridgeMessage,
+  RosbridgeFragmentMessage,
+} from "../types/protocol.js";
+import {
+  isRosbridgeFragmentMessage,
+  isRosbridgeMessage,
+  isRosbridgePngMessage,
+} from "../types/protocol.js";
+import * as ws from "ws";
```
EzraBrooks: I think you need to make this a type import of just the type you need so that it gets dropped by Vite for the production browser build

EzraBrooks:
Since this is nearly there, I'm gonna poke at it, if you don't mind. really trying to get 2.0 out the door before I get busy at my new job 😆

EzraBrooks:
oop, except this PR doesn't have "allow changes from maintainers" enabled

EzraBrooks:
Here's a proposed change to allow fully lazy-loading ws. With this, once the tests are adjusted to pass, I think we're good to merge.

[0001-Move-WsWebSocketTransport-to-lazy-loaded-file.patch](https://github.com/user-attachments/files/23532745/0001-Move-WsWebSocketTransport-to-lazy-loaded-file.patch)

douglascayers:
> oop, except this PR doesn't have "allow changes from maintainers" enabled

Sorry about that! I don't even see where I can enable it =/

I'll work on getting the patch applied. Confused why the tests are failing, they all pass locally 🤔

EzraBrooks:
Yeah I think it defaults to being turned off when the source branch is from an organization

EzraBrooks:
> Confused why the tests are failing, they all pass locally

probably need to reinstall deps, since I changed the underlying PNG package

douglascayers:
> > Confused why the tests are failing, they all pass locally
> 
> probably need to reinstall deps, since I changed the underlying PNG package

Yep, thanks, I'm reworking the tests that were mocking pngparse

douglascayers:
Changes pushed, ready for CI to run 😎

douglascayers:
> [Vitest] Unexpected .only modifier.

doh, removing that...

EzraBrooks (approved):
LGTM. will re-run that flaked test. need to figure out what the deal with that is.. it happens like 1% of the time

EzraBrooks:
thanks so much for the contribution and sorry for the back-and-forth!

douglascayers:
Woo hoo! Happy to have contributed, thanks for shepherding this project forward 👏

PR #1066, opened 2025-11-13
Enable `noImplicitAny`

EzraBrooks:
wow, we'd done such a good job up to this point that this wasn't a miserable slog

### DIFF
--- a/src/core/Ros.ts
+++ b/src/core/Ros.ts
@@ -33,6 +33,10 @@ import type {
 } from "./transport/Transport.js";
 import { WebSocketTransportFactory } from "./transport/WebSocketTransportFactory.ts";
 
+interface TypeDefDict {
+  [key: string]: string | string[] | TypeDefDict | TypeDefDict[];
+}
+
 /**
  * Manages connection to the rosbridge server and all interactions with ROS.
  *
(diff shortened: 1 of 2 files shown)


src/core/Ros.ts
```diff
@@ -33,6 +33,10 @@ import type {
 } from "./transport/Transport.js";
 import { WebSocketTransportFactory } from "./transport/WebSocketTransportFactory.ts";
 
+interface TypeDefDict {
+  [key: string]: string | string[] | TypeDefDict | TypeDefDict[];
+}
```
EzraBrooks, line 38: literally wtf is this type lmao

tsconfig.json
```diff
@@ -53,7 +53,7 @@
 
     /* Rules to be enabled */
     "exactOptionalPropertyTypes": false /* Differentiate between undefined and not present when type checking */,
-    "noImplicitAny": false /* Enable error reporting for expressions and declarations with an implied 'any' type. */,
```
bjsowa, line 56: Should probably move it to the `enabled` section but if your next PR is to just replace every rule with `strict: true` then it doesn't really matter.

bjsowa (approved):

tsconfig.json
EzraBrooks: yep, one sec, I realized this and also that as of this PR we can set `strict: true` and remove a bunch of these explicit settings

PR #1067, opened 2025-11-14
Export Rosbridge protocol types

douglascayers:
**Public API Changes**

Makes available the rosbridge protocol types.

**Description**

With the introduction of transports in https://github.com/RobotWebTools/roslibjs/pull/1060, library consumers may want to implement their own transports. When implementing the `ITransport` interface, the `send` and `on('message')` methods use the `RosbridgeMessage` type.

However, currently developers can't import or reference that type because it's not technically exported.

> Namespace '"/<your_app>/node_modules/roslib/dist/RosLib"' has no exported member 'RosbridgeMessage'.ts(2694)

image


EzraBrooks (commented):
Thanks! However I think this actually might be too much detail - a lot of these APIs aren't exposed at the edge. Let me break apart my earlier PR to expose our edge API types to satisfy this issue

EzraBrooks:
Alternatively, so we don't have to wait for review from other maintainers, you could change this PR to do part of what my earlier PR does - enable failing on warnings in typedoc - and just expose all the missing APIs instead of trying to refactor some out like I did!

EzraBrooks:
superseded by #1073

douglascayers:
Thanks

PR #1072, opened 2025-11-14
Refactor: Exhaustive Bridge Protocol

### LINKED ISSUE #1068: Suspicious `bind` call in Action class
I was going through some of the updated roslib code to add some more dynamic typing (great job on the rest of the refactor btw, sorry I couldn't add more stuff) and I noticed this snippet in `Action#advertise`:

```ts
    this.ros.on(this.name, (msg) => {
      if (isRosbridgeSendActionGoalMessage(msg)) {
        this.#executeAction.bind(this);
      } else {
        throw new Error(
          "Received unrelated message on Action server event stream!",
        );
      }
    });
```

Isn't this just a no-op? I noticed that it changed in d37cf8d1458e002c8b5ef9a693609113752b5863 to satisfy some ESLint conditions, but this change seems to do nothing.  
The old version actually passed the bound function to the event, this one just does nothing. I assume it's meant to be called?

@EzraBrooks

### LINKED ISSUE #1069: Inconsistent impl of Rosbridge Protocol in `Service` class
`Service#callService` sends the op message
```ts
const call = {
      op: "call_service",
      id: serviceCallId,
      service: this.name,
      type: this.serviceType,
      args: request,
      timeout: timeout,
};
```

which adds in `type: this.serviceType`.  
As far as I'm aware, the Rosbridge protocol doesn't specify this field when calling a service ([ref](https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/ROSBRIDGE_PROTOCOL.md#338-call-service))

drewhoener:
**Public API Changes**
None

**Description**
While the typing work was extensive to represent the RosBridge protocol, there were some things that were overlooked in the migration that meant that the type checking both for the developers of this library, and the types delivered to developers working with the lib weren't exhaustive.  
This PR addresses this by refactoring the protocol files:
* Add the documentation from the ROSBridge protocol doc to aid developers which took forever but is worth it (at least for me)
* Create union types both for subsections of the protocol, but also every possible operation. This means that type checking can be narrowed just by checking the 'op' field without having to have a guard method for each operation.
* Fixes some discrepancies that arose as a result of the non-exhaustive checking.
* Allows a distinction between outgoing and incoming service call messages, which takes care of the TODO left by Ezra w.r.t being deep partial on the way out but required on the way in.
* Allows inference on emitter types for special name topic strings like service call and action callback messages.

Apologies for the name changes, I had this left over from another PR I didn't get to complete and the scheme was completely different from yours.

This PR also does some housekeeping on complex methods with large Cyclomatic Complexity, making them easier to read and maintain. In particular, `Service#advertise` and `advertiseAsync` have a common parent function now, joining the common parts together.

I also split out some of the longer that were declared right in function parameters, which makes them harder to read and follow when you have a large object in the constructor.

As always, thanks for the consideration and happy to address any comments :)

Fixes #1068 
Fixes #1069

### DIFF
--- a/src/types/emitted_events.ts
+++ b/src/types/emitted_events.ts
@@ -1,0 +1,38 @@
+import type { TransportEvent } from "../core/transport/Transport.js";
+import type {
+  BridgeProtoOpKey,
+  AnyActionOp,
+  AnyServiceOp,
+} from "./protocol.js";
+
+export type TransportEventString = "open" | "close" | "error";
+export type TransportEvents = Record<
+  TransportEventString,
+  [event: TransportEvent]
+>;
+
+export type ActionIdString = `send_action_goal:${string}:${string}`;
+export type ServiceCallIdString = `call_service:${string}:${string}`;
+export type TopicKey = Exclude<
+  string,
+  TransportEventString | ActionIdString | ServiceCallIdString | BridgeProtoOpKey
+>;
+
+// Event strings specific to actions
+export type RosActionEvents = Record<
+  ActionIdString,
+  // Allow the action class to narrow the type itself.
+  [message: AnyActionOp<never, never, never>]
+>;
+
+export type RosServiceEvents = Record<
+  ServiceCallIdString,
+  // Allow the action class to narrow the type itself.
+  [message: AnyServiceOp<never, never>]
+>;
+
+export type RosEventTypes = TransportEvents &
+  RosActionEvents & // Action Events
+  RosServiceEvents &
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  Record<TopicKey, [message: any]>; // Service Calls & Topic messages, need to be `any` to be coerced by implementers
--- a/src/util/type-utils.ts
+++ b/src/util/type-utils.ts
@@ -1,0 +1,12 @@
+/**
+ * Type-safe wrapper for Object.hasOwn.
+ * Asserts on return that the object has the property, allowing it to be used without type complaints.
+ * @param obj The object to check.
+ * @param prop The property to check for.
+ */
+export function hasOwn<X extends object, Y extends PropertyKey>(
+  obj: X,
+  prop: Y,
+): obj is X & Record<Y, unknown> {
+  return Object.hasOwn(obj, prop);
+}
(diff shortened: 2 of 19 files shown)


src/util/type-utils.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Type-safe wrapper for Object.hasOwn.
+ * Asserts on return that the object has the property, allowing it to be used without type complaints.
+ * @param obj The object to check.
+ * @param prop The property to check for.
+ */
+export function hasOwn<X extends object, Y extends PropertyKey>(
```
EzraBrooks, line 7: Did you know the `in` operator already does this type guard?

src/types/emitted_events.ts
```diff
@@ -0,0 +1,38 @@
...
+  ServiceCallIdString,
+  // Allow the action class to narrow the type itself.
+  [message: AnyServiceOp<never, never>]
+>;
+
+export type RosEventTypes = TransportEvents &
+  RosActionEvents & // Action Events
+  RosServiceEvents &
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  Record<TopicKey, [message: any]>; // Service Calls & Topic messages, need to be `any` to be coerced by implementers
```
EzraBrooks, line 38: Please use `unknown` instead of `any`.

EzraBrooks (commented):
Could you break this down into more atomic PRs? I'm finding it hard to determine which changes are related to which.

EzraBrooks:
I had a brain blast on how to (tediously, but thoroughly) strictly enforce the message types everywhere. I'll put up a PR for your review

src/util/type-utils.ts
drewhoener: `in` will also check the prototype chain which isn't always something you want. Though for the places I use it I suppose I could just use `in`

EzraBrooks: Yep - however, given that we have no inheritance other than from `Object` on these messages, I think it's fine

EzraBrooks:
please let me know what you think about #1075 - it caught the issue you reported with Service, plus a couple others

drewhoener:
Superseded by #1075, I'll open different PRs to address my other maintainability changes

PR #1075, opened 2025-11-14
Use enormous union type to make message types stricter

### LINKED ISSUE #1069: Inconsistent impl of Rosbridge Protocol in `Service` class

EzraBrooks:
Fixes #1069 and a couple other things we hadn't caught yet. (also, validates my instinct that `= unknown` for type args is a lot less painful for static typing than `= undefined`, because I had to change some of them from the latter to the former for this to work)


EzraBrooks:
fry squinting dot gif

why did this build locally

EzraBrooks:
of course right after I tell @douglascayers it's not necessary to expose all these types I go and make the types more exhaustive and it turns out I was wrong 😆

drewhoener:
This basically accomplishes a bunch of what I was doing in my other PR so it's fine by me :)

Couple things for consideration maybe?
* This still doesn't solve the TODO for the deep partial on outgoing + required on incoming for service call operations
* Part of my rationale with the discriminated union was to avoid having to maintain a guard function for each op, those can probably just go away now, right?
* Missing types for compression on RosbridgeCallServiceMessage

This also doesn't provide typing for service or action events emitted by the Ros client. 
Maybe I could use this as a jumping off point and build off of yours? It accomplishes most of the same stuff

EzraBrooks:
Yeah, that would be great - I was just about to comment on your PR that I hope you don't take my jumping in and fixing these issues in a slightly different way as my not appreciating your contribution. There are definitely some other great fixes and doc improvements in your PR!

drewhoener:
Oh no offense taken at all! The renaming was mostly just because I already had the type file from a previous PR I didn't get to submit, and like you said -- it's tedious lol.  
I can work in my stuff no problem though.

src/core/Action.ts
```diff
@@ -252,11 +247,11 @@ export default class Action<
    */
   setFailed(id: string) {
     const call = {
-      op: "action_result",
+      op: "action_result" as const,
```
douglascayers: curious, why was `as const` necessary?

douglascayers (approved):
LGTM 🫡

src/core/Action.ts
EzraBrooks: oop. thanks for catching that. moving this inline to the invocation of `callOnConnection` like I did elsewhere fixes the problem in a more obvious way. using `as const` helped TypeScript understand which message type it was, but I should've either used a type annotation or passed it directly into the parameter

EzraBrooks: moved it inline

MatthijsBurgh (approved):

PR #1081, opened 2025-11-17
Rearrange GitHub Actions to allow publishing to NPM

EzraBrooks:
also temporarily set the package.json version to 2.0.0-alpha1 so I don't accidentally clobber the 2.0.0 version while testing

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,29 @@
+name: CD
+
+on:
+  release:
+    types: [released]
+
+permissions:
+  contents: read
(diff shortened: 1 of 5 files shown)


.github/workflows/cd.yml
EzraBrooks, line 1: This PR is mostly just reorganizing things so I could create a dedicated workflow for Trusted Publishing to NPM so we don't need to use an NPM token secret.

EzraBrooks:
once this and the follow-up PR #1082 are in, I think we're good to cut a pre-release of 2.0.

bjsowa (approved):

PR #1082, opened 2025-11-17
Use `changelog-maker` for `CHANGELOG.md`

EzraBrooks:
Adopts the tool that the Node.js team uses to generate their `CHANGELOG.md` file, adds a first cut at an automation to add the new changelog inline with the existing content (not perfect yet, but good enough to ship 2.0)


EzraBrooks:
Looks like a big addition to the dependencies, but that's just because we hadn't actually captured our previous changelog creator as a dependency.

bjsowa (approved):

PR #1086, opened 2025-11-17
Remove double build in CI/CD

EzraBrooks:
We run `build` during `npm prepare`, which means it's already run implicitly during the `npm ci` step that already exists in these jobs.


EzraBrooks:
meh. it's actually kind of nice for it to fail discretely.

PR #1093, opened 2025-11-17
create npm workspace for monorepo

EzraBrooks:
Creates a root-level `package.json` that lets us run `ci`/`install`, `build` and `test` on all packages in the repo with one command.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -28,9 +28,9 @@ jobs:
           node-version: "lts/*"
           registry-url: "https://registry.npmjs.org"
       - run: npm ci
-      - run: npm run build
-      - run: npm publish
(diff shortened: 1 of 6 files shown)


drewhoener:
Out of curiousity (since the topic is open), have you considered pnpm for the project? It has some nice features to build off of base npm workspaces and is usually a drop-in replacement.  
You get some better checks from the stricter dependency management and it's good about deduplicating packages in node_modules, which for a monorepo can save a lot on space

EzraBrooks:
PNPM is my package manager of choice, yes. My current plan is to gather up all the robotwebtools JS libraries, which are NPM projects, into a monorepo, and then convert it to PNPM all at once afterward

bjsowa (approved):

PR #1096, opened 2025-11-17
Switch from cbor-js to cbor2

### LINKED ISSUE #1095: `cbor-js` is not ESM-compatible
sigh. at least switching to `cbor2` shouldn't be a user-facing change.

EzraBrooks:
**Public API Changes**

64-bit integers will now properly be represented as BigInts.

**Description**

Closes #1095. I was hoping to leave this be, but it's being annoying while I'm testing pure-ESM use cases for roslibjs. At least it deletes a bunch of code and fixes the loss-of-precision issue we used to warn about when using CBOR.


src/core/transport/Transport.ts
```diff
@@ -221,7 +220,7 @@ export abstract class AbstractTransport
    * It is one technique for compressing JSON data.
    */
   private handleCborMessage(cbor: ArrayBuffer) {
-    const data: unknown = CBOR.decode(cbor, typedArrayTagger);
+    const data: unknown = decode(new Uint8Array(cbor));
```
EzraBrooks, line 223: Uint8Arrays are views on ArrayBuffers, so this (thankfully) doesn't do any copying of potentially large data.

drewhoener:
@EzraBrooks Feel free to pull from #926 for this if it'll make it easier, I had a commit in there to replace all the cbor stuff even though it didn't end up making it in.

19cc0d95846a8855768185481246a35bfd9262b8

EzraBrooks:
Thanks Drew! I think this is good to go already, I wished I'd remembered to look at where you'd done it

EzraBrooks:
Hey @bjsowa @sea-bass could one of you take a look at this? I got a request internally to SoftServe to update the browser-based roslibjs examples (which I was going to wait on til after the release), and this and my following PR are causing compatibility issues

MatthijsBurgh (approved):

PR #1102, opened 2025-11-18
Make examples runnable from package.json

EzraBrooks:
- Make the examples work for 2.0
- add a script to run them
- add an index to the examples


sea-bass (approved):
The examples index page is a nice touch :shipit:

PR #1103, opened 2025-11-18
Fix: Check status code in Action.sendGoal instead of result field

### LINKED ISSUE #889: Ros2 Action client does not properly handle status of action response.
The current implementation of Action's `sendGoal` does not look into the response's status code and instead relies on result. This is error-prone and doesn't properly handle STATUS_CANCELLED in particular.

ElshadHu:
**Public API Changes**

**Description**

I fixed a bug in `Action.sendGoal` where the method was checking the `result` field instead of the `status` field to determine action success/failure. This caused `STATUS_CANCELED` and other status codes to be handled incorrectly.

After reading the documentation, I realized that action result messages contain two fields:
- `result` : A simplified success/failure indicator  
- `status` : The authoritative status 

I updated the `sendGoal` method in `src/core/Action.ts` to check `message.status` against `GoalStatus.STATUS_SUCCEEDED` as the primary indicator of success. For all other status values, I route to the failure callback with descriptive error messages based on the specific status code.
```typescript
// Check status code instead of result field
const status = message.status as GoalStatus;

if (status === GoalStatus.STATUS_SUCCEEDED) {
  resultCallback(message.values as TResult);
} else {
  // Handle failures with appropriate error messages
  failedCallback(errorMessage);
}
```

### Testing

I added comprehensive unit tests in `test/actionSendgoal.test.ts` covering:
- GoalStatus values (SUCCEEDED, CANCELED, ABORTED, CANCELING, UNKNOWN)
- Feedback message handling
- Edge case where `result` and `status` fields contradict each other

### References

- [Rosbridge Protocol - Action Result](https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/ROSBRIDGE_PROTOCOL.md#3315-action-result)
- [ROS 2 GoalStatus](https://docs.ros2.org/latest/api/action_msgs/msg/GoalStatus.html)

If there is any issue, please let me know.

Fixes #889

### DIFF
--- a/src/core/Action.ts
+++ b/src/core/Action.ts
@@ -14,34 +14,6 @@ import {
 import type Ros from "./Ros.ts";
 import { v4 as uuidv4 } from "uuid";
 
-export class GoalError extends Error {
-  constructor(status: GoalStatus | number | string, baseError?: unknown) {
-    const base = typeof baseError === "string" ? baseError : "";
-    let message: string;
-
-    switch (status) {
-      case GoalStatus.STATUS_CANCELED:
-        message = `Action was canceled${base ? `: ${base}` : ""}`;
-        break;
-      case GoalStatus.STATUS_ABORTED:
-        message = `Action was aborted${base ? `: ${base}` : ""}`;
-        break;
-      case GoalStatus.STATUS_CANCELING:
-        message = `Action is canceling${base ? `: ${base}` : ""}`;
-        break;
-      case GoalStatus.STATUS_UNKNOWN:
-        message = `Action status unknown${base ? `: ${base}` : ""}`;
-        break;
-      default:
-        message = `Action failed with status ${String(status)}${base ? `: ${base}` : ""}`;
-    }
-
-    super(message);
-    this.name = "GoalError";
-    Object.setPrototypeOf(this, GoalError.prototype);
-  }
-}
-
 /**
  * A ROS 2 action client.
  */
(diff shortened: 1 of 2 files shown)


EzraBrooks:
@sea-bass since you're more familiar with the protocol than I am, can you give this a functional review pass before I give it a typescript review pass?

sea-bass (commented):
Added functionality looks good, but I think we probably still want to leave in a case to catch where `message.result` is `false`, implying something may have gone wrong in the backend implementation?

ElshadHu:
> Added functionality looks good, but I think we probably still want to leave in a case to catch where `message.result` is `false`, implying something may have gone wrong in the backend implementation?

Thanks for the feedback, in that case do you want me to handle `message.result` ? I mean, basically combining the way it is in the codebase and my functionality ?

sea-bass (approved):
LGTM -- over to you, @EzraBrooks

src/core/Action.ts
```diff
@@ -69,19 +69,41 @@ export default class Action<
...
+        const status = message.status as GoalStatus;
+
+        // Check status code instead of result field to properly handle STATUS_CANCELED
+        if (status === GoalStatus.STATUS_SUCCEEDED && message.result) {
           resultCallback(message.values);
+        } else {
+          const baseError =
+            typeof message.values === "string" ? message.values : "";
+
+          let errorMessage: string;
```
EzraBrooks: What if we made a `class GoalError extends Error` that did this template interpolation in its constructor and then we called `String()` on the error before sending it to the `failedCallback`? That way in the future if the error handling here changes to include a `throw`, we already have an error type for it.

EzraBrooks (commented):
LGTM with one future proofing nit

src/core/Action.ts
ElshadHu: Is it okay to put  `GoalError`  on top of `Action.ts` ? To be honest i did not find any file that contains custom error classes.

EzraBrooks: Yep, that's fine!

EzraBrooks:
@ElshadHu do you mind if I push a commit to get it a bit closer to what I intended to guide you to, so we don't have to go back and forth?

ElshadHu:
> @ElshadHu do you mind if I push a commit to get it a bit closer to what I intended to guide you to, so we don't have to go back and forth?

Sure

EzraBrooks:
Thanks! if my change looks good to you (and I didn't break the tests 😆) I'm happy to merge this now.

ElshadHu:
> Thanks! if my change looks good to you (and I didn't break the tests 😆) I'm happy to merge this now.

Thanks for the help,  It seems perfect now.

EzraBrooks:
jazzy and kilted are broken (#1104) so we're just looking for humble here.

EzraBrooks (approved):
Thanks for the contribution!

ElshadHu:
> Thanks for the contribution!

It is my pleasure to contribute to such a nice repo.

PR #1105, opened 2025-11-19
Make `isConnected` a getter

EzraBrooks:
**Public API Changes**

Reverts the `isConnected` API to looking like a property, while preserving it as a function in case we need to add extra logic later.

**Description**

^


bjsowa:
@Krzemien97 FYI

PR #1106, opened 2025-11-19
Add delay_between_messages argument to launch file

bjsowa:
**Public API Changes**
None

**Description**
Testing weather this fixes CI


bjsowa:
@MatthijsBurgh @EzraBrooks jazzy and kilted fail the param set test due to another bug I fixed in RobotWebTools/rosbridge_suite#1133

Also @MatthijsBurgh, the workaround for humble will make the tests fail upon the next humble sync.

EzraBrooks:
I think we should disable jazzy and kilted from the matrix until next sync. Failing CI jobs that can't be fixed are bad for signal-to-noise ratio.

bjsowa:
> I think we should disable jazzy and kilted from the matrix until next sync. Failing CI jobs that can't be fixed are bad for signal-to-noise ratio.

I'm fine with that

Issue #1062, opened 2025-11-13
Migrate examples to their own package, re-architect their tests


EzraBrooks:
(after monorepo migration)

The "example tests" seem to be manually-synchronized with the contents of the "examples", which is not a maintainable strategy.

Extract the examples to their own package, then use Playwright to actually execute the examples and click their buttons and such to ensure that they work.

PR #1110, opened 2025-11-20
Move examples to their own package

EzraBrooks:
step 1 of #1062. working on figuring out how to make the rosbridge test container launcher portable to this package and others so I can deduplicate the example tests.


packages/roslib-examples/package.json
```diff
@@ -0,0 +1,38 @@
...
+    }
+  ],
+  "scripts": {
+    "start": "http-server .. -o /roslib-examples/src/index.html"
+  },
+  "devDependencies": {
+    "http-server": "^14.1.1"
+  },
+  "dependencies": {
+    "roslib": "^2.0.0"
```
EzraBrooks, line 36: roslib will automatically be symlinked from the other directory because it's in the same workspace, this specification only affects the distributed package (if we distribute this package)

bjsowa (approved):

README.md
```diff
@@ -6,3 +6,4 @@ See the READMEs of individual packages for more information.
 ## Packages
 
 - [roslibjs](./packages/roslib/README.md)
+- [roslibjs examples](./packages/roslib-examples/README.md)
```
bjsowa, line 9: This confuses me. Is it `roslib` or `roslibjs` ? Where does the difference come from?

EzraBrooks: The package is actually called `roslib`, I guess because historically the "JS" part was redundant because it was the only roslib and it was being published as a javascript package?

We can decide to change the name later when we eventually adopt an organization namespace on NPM since that will require changing the name anyway

PR #1117, opened 2025-11-25
fix action example feedback message (for example_interfaces/Fibonacci type)

Spir0u:
**Public API Changes**
None

**Description**
The fibonacci type was chaged from action_tutorials_interfaces/Fibonacci to example_interfaces/Fibonacci in #955, which change the feedback name


EzraBrooks (approved):
Thanks. Hoping to get some integration tests in for these to avoid problems like this in the future in #1111 but running into some logistical hurdles at the moment

MatthijsBurgh (approved):

Issue #1120, opened 2025-12-11
GoalStatus only exported as type


mcschwa:
Hi, GoalStatus currently is only exported as type. So at typescript runtime I can't use its values. This means I have to duplicate it locally. I'm not sure if this is intended or not.

Here's a code snippet showing a composable I made that inits as STATUS_UNKNOWN. But this obviously doesn't work

```ts
import { GoalStatus } from "roslib";
import { ref } from "vue";

import useChargeStatusTopic from "@/composables/topics/charge/use-charge-status-topic";


const chargeStatus = ref<GoalStatus>(GoalStatus.STATUS_UNKNOWN); <- TS1362: GoalStatus cannot be used as a value because it was exported using export type

export default function useChargeStatus() {
  const topic = useChargeStatusTopic();

  topic.subscribe((msg) => {
    chargeStatus.value = msg.status_list.at(-1).status;
  });

  return chargeStatus;
}
```

and thus I have to create my own type definition.

I could technically make it undefined too, but I'm not interested in that because I'd much rather just use `STATUS_UNKNOWN`.

If unintended and required to be fixed / changed I'd be willing to create a PR.

PR #1121, opened 2025-12-11
chore: export GoalStatus

ikwilnaarhuisman:
**Public API Changes**
Made the `GoalStatus`' values public, so they may be initialized externally.

**Description**
This pull request allows for the Goal Status to be fully used externally, instead of only the type.

https://github.com/RobotWebTools/roslibjs/issues/1120


EzraBrooks (approved):

ikwilnaarhuisman:
Awesome, this will greatly improve working with the library.

PR #1128, opened 2025-12-15
chore: update cancelGoal to handle optional id parameter

ikwilnaarhuisman:
The Ros1 ActionClient has the functionality that allows all goals to be canceled that were associated with the ActionClient. I thought this was very nice and useful, but missing in the Ros2 implementation of the Action / ActionClient.

Because I'm working on a project that allows for multiple users / clients to send goals and cancel those goals I don't have the correct goalID nor the means to really get it. Hence why I think this is useful.


EzraBrooks (changes requested):
I don't think we should conflate "cancel a single goal" with "cancel all goals". However, this could be its own function `cancelAllGoals`, which I'd be happy to merge if you add an integration test showing it does work in ROS 2.

ikwilnaarhuisman:
closed in favour of https://github.com/RobotWebTools/roslibjs/pull/1130 because I'm not the git master I wish to be.

PR #1130, opened 2025-12-16
Add `Action.cancelAllGoals()`

ikwilnaarhuisman:
The Ros1 ActionClient has the functionality that allows all goals to be canceled that were associated with the ActionClient. I thought this was very nice and useful, but missing in the Ros2 implementation of the Action / ActionClient.

Because I'm working on a project that allows for multiple users / clients to send goals and cancel those goals I don't have the correct goalID nor the means to really get it. Hence why I think this is useful.

**Changes:**
- Added cancelAllGoals -> cancels all goals for an action.
- Added integration tests
  - Added ROS2 fibonacci integration test based on the ROS1 fibonacci integration test. (There's an issue in the fibonacci server though; it doesn't return the correct order. But this doesn't change anything for the implementation of the test).
  - Added an integration test that verifies (based on fibonacci Action) that cancelAllGoals works, by checking whether failedCallback was invoked and not resultCallback.

### DIFF
--- a/packages/roslib/src/core/Action.ts
+++ b/packages/roslib/src/core/Action.ts
@@ -135,6 +135,17 @@ export default class Action<
   }
 
   /**
+   * Cancels all action goals.
+   */
+  cancelAllGoals() {
+    this.ros.callOnConnection({
+      op: "call_service",
+      service: `${this.name}/_action/cancel_goal`,
+      args: {},
+    });
+  }
+
+  /**
    * Advertise the action. This turns the Action object from a client
    * into a server. The callback will be called with every goal sent to this action.
    *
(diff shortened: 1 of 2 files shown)


ikwilnaarhuisman:
@EzraBrooks Because `process.env["ROS_VERSION"] === "2"` didn't work for skipping specific tests I chose to use `process.env["ROS_DISTRO"] === "humble"` because this is the only ros2 version the pipeline uses. I think in the future we may want to fix / set `process.env["ROS_VERSION"]` because of jazzy.

EzraBrooks (approved):
Awesome, thanks!

ikwilnaarhuisman:
No problem, happy to help! We at Lowpad are very thankful for this repo!

Issue #1145, opened 2026-01-20
Missing functionality in actions


ikwilnaarhuisman:
**Actions** in Ros2 are great and amazingly implemented in Roslibjs, but I'm missing functionality (I think, do correct me if I'm wrong). 

What I think we're missing:

The ability to directly subscribe on the `feedback` of an `action`, which is always a `topic` under the **action**, right?

And

The ability to directly subscribe on the `status` of an `action`; which (the same as feedback) is always a topic under the **action** right?

So my question is; was it intended that Action doesn't have a direct subscribe on feedback and or status? Because right now I'm finding myself to be creating the same topics over and over for actions. 

For clarity: The reason I want this functionality is because we may have multiple instances of our application running so when user **A** sends a goal, user **B** needs to know this too. Hence we use the status and feedback of an action.
-> This is also why I requested this change: https://github.com/RobotWebTools/roslibjs/pull/1130

ikwilnaarhuisman:
Oh and I do understand that the message type may be different for **feedback** topics under different actions, however I'm not sure how much that matters.

PR #1146, opened 2026-01-21
experimental: add getStatusTopic method to retrieve action status

ikwilnaarhuisman:
https://github.com/RobotWebTools/roslibjs/issues/1145 <- issue (which this PR fixes 50% of) 

AFAIK the status of an action (aka the topic) is always defined in the same way: `/{ACTION_NAME}/_action/status` with message type `action_msgs/msg/GoalStatusArray`. So instead of having to create the topic yourself, you can just get it directly from the action.

This PR is **experimental** because I doubt that this is acceptable _yet_, but it works and gives a better idea than my vague description of the "missing functionality"

### DIFF
--- a/packages/roslib/src/core/Action.ts
+++ b/packages/roslib/src/core/Action.ts
@@ -3,6 +3,7 @@
  * @author Sebastian Castro - sebastian.castro@picknik.ai
  */
 
+import { actionlib_msgs } from "../types/actionlib_msgs.ts";
 import { GoalStatus } from "./GoalStatus.ts";
 import type { RosbridgeSendActionGoalMessage } from "../types/protocol.ts";
 import {
@@ -13,6 +14,8 @@ import {
 } from "../types/protocol.ts";
 import type Ros from "./Ros.ts";
 import { v4 as uuidv4 } from "uuid";
+import Topic from "./Topic.ts";
+import GoalStatusArray = actionlib_msgs.GoalStatusArray;
 
 class GoalError extends Error {
   override name = "GoalError";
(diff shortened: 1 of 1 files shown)


ikwilnaarhuisman:
I want to achieve the same functionality for the feedback topic, but I'm not sure how considering the message type differs:
```typescript

  getFeedbackTopic() {
    return new Topic<TFeedback>({
      ros: this.ros,
      name: `${this.name}/_action/feedback`,
      messageType: "", // <<<< unknown! doesn't even work.
    });
  }

// OR

  getFeedbackTopic(messageType: string) {
    return new Topic<TFeedback>({
      ros: this.ros,
      name: `${this.name}/_action/feedback`,
      messageType: messageType, // works, but feels dirty.
    });
  }
```

ikwilnaarhuisman:
@EzraBrooks What's your opinion on this?

EzraBrooks (commented):
Personally I'm philosophically opposed to exposing the underlying `_action` namespaced topics in RWT because they're supposed to be ROS implementation details that clients do not mess with.

If there's a compelling reason to do this (which I'm not sure there is - I think perhaps you are operating outside the intended usage of Actions :wink:), I think we should expose it at the protocol level in rosbridge/rosapi (cc @sea-bass @bjsowa) rather than adding a function to `Action` that returns a `Topic` - I really appreciate that our ROS 2 Action implementation in our client libraries can be disentangled from `Topic`, `Service`, etc. unlike the original actionlib implementation.

If it were done at the protocol level, then the roslibjs implementation could emit a `status` event on the `Action` class, instead of returning a separate subscribable `Topic`.

PR #1169, opened 2026-03-06
feat: Add Promise-based `callServiceAsync` wrapper for async/await support

ikwilnaarhuisman:
**Public API Changes**
Added `callServiceAsync` with a promise return. Which allows for cleaner usage.

for example: (response based on default srv implementation from:  (https://github.com/ros2/common_interfaces/tree/rolling/std_srvs/srv)
```ts
// Originally:
service.callService(
  request,
  (response) => {
    if (response.success) {
      console.log("request successful");
    } else {
      console.error("request failed");
    }
  },
  (failedCallbackString) => {
    console.error("request failed");
  }
);

// Which now may be replaced with:
try {
  const response = await service.callServiceAsync(request);
  
  if (response.success) {
    console.log("request successful");
  } else {
    console.error("request failed");
  }
} catch (error) {
  console.error("request failed");
}

```

In my opinion this is a lot more readable and a bit more normalized in the "typescript / javascript" world.

### DIFF
--- a/packages/roslib/src/core/Service.ts
+++ b/packages/roslib/src/core/Service.ts
@@ -96,6 +96,29 @@ export default class Service<
       timeout: timeout,
     });
   }
+
+  /**
+   * Wrapper of callService that returns a modern Promise-based interface for use with async/await.
+   * @see callService
+   * @param request - The service request to send.
+   * @param [timeout] - Optional timeout, in seconds, for the service call. A non-positive value means no timeout.
+   *                             If not provided, the rosbridge server will use its default value.
+   */
+  callServiceAsync(request: TRequest, timeout?: number): Promise<TResponse> {
+    return new Promise<TResponse>((resolve, reject) => {
+      this.callService(
+        request,
+        (res) => {
+          resolve(res);
+        },
+        (err) => {
+          reject(new Error(err));
+        },
+        timeout,
+      );
+    });
+  }
+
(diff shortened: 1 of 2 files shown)


ikwilnaarhuisman:
added a unit test (or integration test... whatever you'd like to call it) based on the https://docs.ros.org/en/hydro/api/rosapi/html/srv/GetTime.html 

Which I think is nice considering we'll probably always be using the rosbridge, right?

ikwilnaarhuisman:
Shit! Seems that ros1 does not support the `getTime` service from rosapi, should I make the unit test limited to ROS_VERSION 2. Considering everything's sourced well you can easily get that environment variable. LMK!

EzraBrooks (commented):
This is a great contribution! I had done this as a wrapper of `callService` in my previous project, having it in the core library will be very useful.

For the test, what about creating a service server in the test code itself and then calling that?

ikwilnaarhuisman:
> This is a great contribution! I had done this as a wrapper of `callService` in my previous project, having it in the core library will be very useful.

 lmao: image


> For the test, what about creating a service server in the test code itself and then calling that?

Sure! I'll take a gander at how it's done with the other unit tests and push ASAP.

ikwilnaarhuisman:
Hooray

EzraBrooks (approved):

Issue #1170, opened 2026-03-06
promise based methods


ikwilnaarhuisman:
in PR https://github.com/RobotWebTools/roslibjs/pull/1169 I added `callServiceAsync` to `Service` which is just a promise based wrapper around the regular `callRosService`.

Do we want this for other things too? 

Say `sendGoal` for example for `Action`?

MatthijsBurgh:
I think this is a good idea.

Though is it possible to make the sync function wrap the async function? Instead of the async function wrapping the sync function.

ikwilnaarhuisman:
Yeah so we can deprecate the sync ones in a 3.x? 

Sounds like a good idea.

ikwilnaarhuisman:
in https://github.com/RobotWebTools/roslibjs/pull/1166 @cn1001wang said that he added promise based `Params` and `Services` . 

It looks pretty good: https://github.com/cn1001wang/roslib-ts/blob/v1.0.0/src/Param.ts

MatthijsBurgh:
It would be really nice to have more async functions.

PR #1173, opened 2026-03-10
chore: make callService wrap callServiceAsync instead of around

ikwilnaarhuisman:
**Public API Changes**
Made the `callService` method wrap `callServiceAsync` instead of around.

**Description**
@MatthijsBurgh (https://github.com/RobotWebTools/roslibjs/issues/1170#issuecomment-4029131063) came with this idea and I think it's a lot more logical for the future of this package.


EzraBrooks (approved):
LGTM

